PR #6, opened 2023-05-25
Windows support, solidified.

aremmell:
Windows has been more or less an afterthought since the beginning. I recently decided to see if I could create a Visual Studio solution, and went down the rabbit hole...

- Created Visual Studio 2022 solution and projects in the `msvc` folder. arm64 and x86_64 configurations are implemented and tested (builds as a static library; no DLL support)
- Fixed bugs (that not only affected Windows)
- Implemented implementation/extension specific wrappers for things like `fopen_s`, `strncpy_s`, etc.
    - Consolidated Windows' `__STDC_SECURE_LIB__` and libc's `__STDC_LIB_EXT1__` to accomplish this.
- Vastly improved test rig (organization, deduping)
- Improved stdio perf by around 66% through utilizing binary search when looking up things like text styling or error messages.
- Squashed many warnings from `cl`'s highest warning level.
- Added new error definitions for NULL pointer args, and invalid args (w/ associated verifier functions)
- Added support for **FreeBSD** _(and, in theory, NetBSD, OpenBSD, etc. but haven't tried anything other than FreeBSD)_
    - One issue: you have to use `gmake` on FreeBSD or the build will fail. I think there are many good reasons to switch to `CMake` and ditch the Makefile.


aremmell:
Unverified commit [bad280f](https://github.com/aremmell/libsir/pull/6/commits/bad280f13c1375b55ee29d612aa7e6bfca62df6e)

Issue #12, opened 2023-06-25
Potential leak of memory pointed to by 'filename' in sirfilesystem.c:250:16 (branch: syslog)


johnsonjh:
per Clang Analyzer:

![Screenshot 2023-06-24 at 23-55-06 sirfilesystem c](https://github.com/aremmell/libsir/assets/61629094/6dad1661-cc7b-4e4d-a526-7d13ad5642a4)

johnsonjh:
Fixed in syslog branch

Issue #17, opened 2023-06-25
Use of memory after it is freed in sirfilesystem.c:244:5 (branch: syslog)


johnsonjh:
per Clang Analyzer:

![Screenshot 2023-06-25 at 00-01-30 sirfilesystem c](https://github.com/aremmell/libsir/assets/61629094/e1aedf9a-44ca-4783-a198-d8278fe673fe)

aremmell:
```c
char* _sir_getappbasename(void) {
    char* filename = _sir_getappfilename();
    if (!_sir_validstr(filename))
        return NULL;
```

johnsonjh:
I'll fix later (not on my phone)

johnsonjh:
Fixed in syslog branch

Issue #16, opened 2023-06-25
Use of memory after it is freed in sirfilecache.c:113:5 (branch: syslog)


johnsonjh:
per Clang Analyzer:

![Screenshot 2023-06-25 at 00-00-26 sirfilecache c](https://github.com/aremmell/libsir/assets/61629094/9a09080c-1971-48de-a6f9-07571a8dce7d)

johnsonjh:
Fixed in syslog branch

PR #39, opened 2023-06-25
Fixes: Closes issues #12, #16, #17

johnsonjh:
* Closes:
  *  #12
  *  #16
  *  #17

> Did this on my phone, please review and make sure I'm not being dumb.


johnsonjh:
Should be good to go

sirfilecache.c
aremmell, line 1: Here's the trick: _sir_safefree sets the pointer to NULL after it frees; so this return NULL is redundant. I think clang-analyzer missed that.

sirfilesystem.c
aremmell, line 1: Well, this isn't the only place `_sir_getappfilename` is called (it's called all over). Why is freeing it inside `if (!resolved)` good enough? I'm confused about that.

tests/tests.c
aremmell, line 1: I believe this should be size_t, no? And how does declaring `err` separately resolve an issue? Sorry, I'm confused about that too.

sirfilecache.c
johnsonjh: Yeah, it's a bit strange ... I just checked Clang and also Purify and they both complain if it isn't there, so I'd say leave it, can't hurt.

tests/tests.c
johnsonjh: The root of the problem was the previous typo which is now fixed, so this shouldn't be needed to pass now; I'll re-simplify it.

sirfilesystem.c
johnsonjh: Reversing this patch on this branch, i.e.:
```diff
diff --git a/sirfilesystem.c b/sirfilesystem.c
index e840bf3..19bb927 100644
--- a/sirfilesystem.c
+++ b/sirfilesystem.c
@@ -239,0 +240 @@ char* _sir_getappfilename(void) {
+       _sir_safefree(buffer);
@@ -248,2 +249 @@ char* _sir_getappbasename(void) {
-    if (!_sir_validstr(filename)) {
-       _sir_safefree(filename);
+    if (!_sir_validstr(filename))
@@ -251 +250,0 @@ char* _sir_getappbasename(void) {
-    }
```

Analzyer reports the following two errors with these paths:

[IMG]
[IMG]

johnsonjh: @aremmell 

All other things you noted are addressed.

johnsonjh:
Rebased; fixes all remaining scan-build errors and passes all tests.

PR #45, opened 2023-06-26
Use Cppi to format directives

johnsonjh:
* Use GNU Cppi to check and format preprocessor directives
* Don't require bash be installed to run autoformat.sh
* Update autoformat.sh to only use available utilities, and warn if any are missing (currently just Cppi)
* Trim unnecessary whitespace, newlines


johnsonjh:
Should be good to go.

@aremmell 

I notice that running your clang-format still re-format some additional files.  I didn't do process those since I assume you left them purposely and are still tweaking the rules and/or haven't gotten around to adding `/* clang-format off */` ... `/* clang-format on */` blocks yet, so I left them as-is.  When you do have it acceptable, you should put clang-format and cppi in a git pre-commit hook, so I don't forget to run it :)

aremmell:
So, a couple things:

1. I don't really intend to use clang-format in the long run, because no matter how I configure it, it ends up doing something I don't want it to do, and then I just have to do extra work.
2. I'm a little overwhelmed with all the submissions (I've never had someone work with me on this), and our styles are very different. Would you be offended if I took your code changes in but in my/libsir's current style and form?
3. I have another repo, https://github.com/aremmell/systest which I built essentially as a testing ground for code I want to add to libsir. What I would like to do is take for example the path retrieval code you wrote and merge it into that, and see how it fares before we stick it in here, if that's all right?

johnsonjh:
1. I'll rework this to only use Cppi.  It's safe - it indents the preprocessor directives only and doesn't change your code.  It also can detect some errors.  

2. Not at all offended, not in the least.  I'm happy that you're willing to consider the contributions.  As I said before, my motivation is (mostly) selfish.  This library does most everything I want and has a license that's compatible with our project.  Whatever I can upstream to you means one less change we need in our vendored version, and makes it easier to track and integrate your future changes.  Also, I'm sorry for sending kind of an onslaught of changes.  I originally forked back in 2018 or 2019 so I had a lot of my own changes (and then some), and I can't reasonably expect you to take them all.

I'll go off on a bit of a tangent here, however.  We use [libDecNumber](https://speleotrove.com/decimal/), a library implementing the IBM General Decimal Arithmetic standard and IEEE-754.  This library was written by IBM's Mike Cowlishaw (of REXX fame), and was generally supported by IBM staff.  It was then donated to the FSF and is an integral part of GCC, providing support for decimal types when the native hardware does not.  So the current library is available under the GPL.  The original version is ICU (MIT variant).  Most but not all patches get contributed to not GCC and the "original" project.  (Still, most  of the changes to this library in GCC are actually done by IBMers.). This leaves us sometimes in the situation where I have to read the changelogs or git commit messages and re-create patches that make it into GCC but without using their actual code, or ask the committer if they are also willing to contribute their changes to the ICU-licensed version.  There are some good alternatives to decNumber now, but, as long as it is an integral part of GCC, we'll likely stick with it, but the whole process is a **huge pain**. So adding modified local forks is something I want to avoid if I can.

3. That works for me.  I can also setup a process that pulls from that (and also libsir) and builds for the platforms that you don't, if that will help.

johnsonjh:
@aremmell This only runs Cppi - no other reformatting right now, see if you like it better?

aremmell:
Same thing here. I just merged master back into syslog...

johnsonjh:
> Same thing here. I just merged master back into syslog...

Easy enough here

aremmell:
> Easy enough here

Oh and dig the new [docs](https://libsir.rml.dev) and let me know what you think. Yes, the purple is a bit much, but so am I.

johnsonjh:
@aremmell should be GTG

johnsonjh:
@aremmell merge away - no code changes; no big formatting changes, just directive indentation and alignments.  Re-running autoformat.sh results in no modifications so if you approve and you merge this it's safe for a pre-commit hook.

Cppi is a great tool, it's helped save us from some tricky if/else problems, since it lints as well as beautifies, and it's error messages are often more descriptive than the C preprocessor.

PR #64, opened 2023-07-06
Re-merged syslog to master and added new commits

johnsonjh:
So, still not sure what happened with master, but I since couldn't figure it out with any tools, so I reset everything back to how it was, did the merge again with fugitive, and then added your newer commits on top.


johnsonjh:
It's not going to be possible for github to merge as-is on the server side because of how it handles conflicts, so I'll force-push the delta here.

johnsonjh:
And that's upsetting GitHub - I'll close and try it another way.

PR #65, opened 2023-07-06
Fix state of master; add support for Cygwin and Haiku.

johnsonjh:
This rewinds master to https://github.com/aremmell/libsir/commit/fe9c43f1d9db15fc51e63edc81dff60f9927de04 and then merges syslog and then replays through 5256a04c4fa00e3ecb176581a08e599a29889ea5, squashed, and regenerated documentation, and PR'd as a delta against *current* master. I am pretty sure I didn't miss anything.

* This re-merging makes all the tests that fail right now on master work again - those are illumos, Solaris, NetBSD, FreeBSD with thread names, Linux/bionic-libc, Linux/musl, Linux/uclibc, and DragonFly BSD. 

* Added support for Cygwin in aa39b9aa5d7f9b9fd3261483aeb29a384b665002, tested on Cygwin 3.5.0 with GCC 11.4.0.

* Added support for Haiku in 1ee748e5dbde78b5e0337b91281b3cbfc5b69ff0, tested on Haiku hrev57134 with GCC 11.2.0 and Clang 12.0.1.

* Simplify `autoformat.sh`.


johnsonjh:
@aremmell This re-merging makes all the tests that fail right now on master work again - those are illumos, Solaris, NetBSD, FreeBSD with thread names, Linux/musl, Linux/uclibc, and DragonFly BSD.

johnsonjh:
FYI - I used Fugitive to do the resolution and had to merge tested by my (local) GitLab, since I can't ever seem to figure out how to make GitHub do the right thing on complex merges, since it starts reporting conflicts on PRs that git itself and GitLab's server side have no problem resolving or require manual interventions.

johnsonjh:
Added support for Cygwin in 705af3ec78c180bc5a56a0916ca18e3ffac1633b.

Tested on Cygwin 3.5.0 with GCC 11.4.0

aremmell:
So, what happened? My merge was wrong? Almost every cherry-pick that I did had conflicts, so I guess I'm not surprised.

I also shouldn't have been working on master that entire time, because some of those merges deleted new code and replaced it with old.

Anyhow, I spent time with scan-build, and fixed the ones it gives me (that are not a red herring)–those null ptr ones have to be wrong. The line before the deref is literally a null pointer check, but I was wrong about the others–the computer is right. On a technicality, it was possible for those to leak, like say, if the filename string returned was jibberish and had a zero as its first byte.

docs/dir_49e56c817e5e54854c35e136979f97ca.html
```diff
@@ -110,16 +118,36 @@
...
+========
   <div class="headertitle"><div class="title">docs Directory Reference</div></div>
 </div><!--header-->
 <div class="contents">
+>>>>>>>> master:docs/dir_49e56c817e5e54854c35e136979f97ca.html
 </div><!-- contents -->
 </div><!-- doc-content -->
 <!-- HTML footer for doxygen 1.9.7-->
 <!-- start footer part -->
 <div id="nav-path" class="navpath"><!-- id is needed for treeview function! -->
```
aremmell, line 120: I need to fix the Doxygen config; I don't want any of these internal structures to end up in the public documentation. There's no point to it–anyone who wants to hack on the library is going to have to learn much more than I can type in a little blurb, and there   are too many blurbs.

sirdefaults.h
```diff
@@ -104,7 +104,7 @@ static const sir_options sir_stderr_def_opts
  */
 static const sir_levels sir_syslog_def_lvls
     = SIRL_NOTICE | SIRL_WARN | SIRL_ERROR | SIRL_CRIT |
-      SIRL_ALERT | SIRL_EMERG;
```
aremmell, line 107: Can you please un-indent this? I like this lined up vertically

sirfilecache.c
```diff
@@ -436,8 +436,8 @@ sirfileid _sir_fcache_add(sirfcache* sfc, const char* path, sir_levels levels,
 
         return &sf->id;
     }
-
```
aremmell, line 439: Similar here; I want that whitespace.

aremmell: Well, at least the one on top.

sirfilesystem.c
```diff
@@ -172,9 +174,19 @@ char* _sir_getappfilename(void) {
...
+#if defined(__linux__)
+# define PROC_SELF "/proc/self/exe"
+#elif defined(__NetBSD__)
+# define PROC_SELF "/proc/curproc/exe"
+#elif defined(__DragonFly__)
+# define PROC_SELF "/proc/curproc/file"
+#elif defined(__SOLARIS__)
+# define PROC_SELF "/proc/self/path/a.out"
+#endif
+
```
aremmell, line 186: I really missed this? Don't let me do merging at 3am.

sirhelpers.h
```diff
@@ -35,7 +35,7 @@
  * Creates an error code that (hopefully) doesn't conflict
  * with any of those defined by the platform.
  */
-# define _sir_mkerror(code) (((uint32_t)((code) & 0x7fff) << 16) | 0x80000000)
+# define _sir_mkerror(code) (((uint32_t)((code)&0x7fff) << 16) | 0x80000000)
```
aremmell, line 38: Please restore spacing around `&`

sirhelpers.h
```diff
@@ -64,32 +64,46 @@ uint16_t _sir_geterrcode(uint32_t err) {
...
 # define _SIR_BEGIN_BIN_SEARCH() do {
-# define _SIR_ITERATE_BIN_SEARCH(comparison) \
-    if (0 == comparison) { \
-        break; \
-    } \
-    \
-    if (_low == _high) \
-        break; \
-    \
+# define _SIR_ITERATE_BIN_SEARCH(comparison)    \
```
aremmell, line 72: I'm ambivalent about this. I almost like the right-align better. Which is your preference?

sirhelpers.h
```diff
@@ -64,32 +64,46 @@ uint16_t _sir_geterrcode(uint32_t err) {
...
+# define _sir_validptrptr(pp) __sir_validptrptr((const void* restrict*)pp, true)
+
+/**
+ * Validates a pointer-to-pointer, pointer,
+ * pointer to function, etc. but ignores whether it's invalid.
+ *
+ * This is necessary due to the fact that ::_sir_validptr will
+ * not accept these types as input.
+ */
+# define _sir_notnull(addr) (NULL != (addr))
```
aremmell, line 105: This should have been deleted; if it's in use anywhere it should not be, and should be replaced with one of:

1. _sir_validptr
2. _sir_validptrptr
3. _sir_validfnptr

sirhelpers.h
```diff
@@ -181,21 +195,19 @@ bool _sir_validptrnofail(const void* restrict p) {
...
 
+/** Validates a pointer-to-pointer and optionally fails if it's invalid. */
 bool __sir_validptrptr(const void* restrict* pp, bool fail);
 
 /** Validates a pointer and fails if it's invalid. */
 #define _sir_validptr(p) __sir_validptr((const void* restrict)p, true)
 
-/** Validates a pointer-to-pointer and fails if it's invalid. */
-#define _sir_validptrptr(pp) __sir_validptrptr((const void* restrict*)pp, true)
-
```
aremmell, line 191: No this belongs here; it replaces _sir_notnull

sirhelpers.h
```diff
@@ -181,21 +195,19 @@ bool _sir_validptrnofail(const void* restrict p) {
...
 /** Validates a pointer-to-function and fails if it's invalid. */
 #define _sir_validfnptr(fnp) __sir_validptrptr((const void* restrict*)&fnp, true)
 
 /** Places a null terminator at the first index in a string buffer. */
 static inline
 void _sir_resetstr(char* str) {
-    str[0] = '\0';
+    str[0] = (char)'\0';
```
aremmell, line 210: I also removed this on purpose; it's a remnant from when I had a typedef for sirchar, and thought some day I might do wide string support.

sirplatform.h
```diff
@@ -124,22 +170,18 @@
...
 # define SIR_MAXHOST 256
 
-# if defined(__GLIBC__)
-#  if (__GLIBC__ >= 2 && __GLIBC_MINOR__ > 19) || \
-     ((__GLIBC__ == 2 && __GLIBC_MINOR__ <= 19) && defined(_BSD_SOURCE))
-#   define __HAVE_UNISTD_READLINK__
-#  endif
-# endif
-
 # if !defined(__WIN__)
```
aremmell, line 173: I swear I specifically recall merging this in. I have no idea what happened there.

sirplatform.h
```diff
@@ -124,22 +170,18 @@
...
 #  include <unistd.h>
 #  include <sys/syscall.h>
 #  include <sys/time.h>
 #  include <strings.h>
 #  include <termios.h>
 #  include <limits.h>
-#  include <fcntl.h>
 #  include <libgen.h>
 #  if defined(__HAVE_ATOMIC_H__) && !defined(__cplusplus)
 #   include <stdatomic.h>
```
aremmell, line 189: Ditto for all this.

tests/tests.c
```diff
@@ -26,30 +26,30 @@
...
-    {"file-cache-sanity",       sirtest_filecachesanity, false, true},
-    {"file-invalid-name",       sirtest_failinvalidfilename, false, true},
+    {"performance",             sirtest_perf,                  false, true},
+    {"thread-race",             sirtest_mthread_race,          false, true},
+    {"exceed-max-buffer-size",  sirtest_exceedmaxsize,         false, true},
+    {"no-output-destination",   sirtest_failnooutputdest,      false, true},
+    {"null-pointers",           sirtest_failnulls,             false, true},
+    {"file-cache-sanity",       sirtest_filecachesanity,       false, true},
+    {"file-invalid-name",       sirtest_failinvalidfilename,   false, true},
     {"file-bad-permissions",    sirtest_failfilebadpermission, false, true},
```
aremmell, line 36: Please restore original spacing

tests/tests.h
```diff
@@ -45,15 +45,15 @@
...
 # define INIT_BASE(var, l_stdout, o_stdout, l_stderr, o_stderr, p_name, init) \
-    sirinit var         = {0}; \
-    var.d_stdout.opts   = (o_stdout) > 0 ? o_stdout : SIRO_DEFAULT; \
-    var.d_stdout.levels = l_stdout; \
-    var.d_stderr.opts   = (o_stderr) > 0 ? o_stderr : SIRO_DEFAULT; \
-    var.d_stderr.levels = l_stderr; \
-    if (_sir_validstrnofail(p_name)) \
-        _sir_strncpy(var.name, SIR_MAXNAME, p_name, SIR_MAXNAME); \
-    bool var##_init = false; \
-    if (init) \
```
aremmell, line 56: This is weird. In the other macro, it left-aligned, now this one is right-aligned. We need to stick with one or the other... Let's go with LEFT.

tests/tests.h
```diff
@@ -328,7 +328,7 @@ static const struct cl_arg {
     } _cl_arg_list[] = {
         {"--perf", "", "Only run the performance measurement test."},
         {"--wait", "", "Wait for a keypress after running test(s) before exiting."},
-        {"--only", "" ULINE("name") " [, name, ...]", "Only run the test(s) specified."},
+        {"--only", ""  ULINE("name") " [, name, ...]", "Only run the test(s) specified."},
```
aremmell, line 331: No space needed here

aremmell (changes requested):
Thanks for the hard work but this one needs some finesse.

sirplatform.h
johnsonjh: I kind of hate git - I think the issue is automatic resolutions (git-rere) stuff, but I guess having Microsoft give us a free platform makes using it worth the pain.  So honestly it probably wasn't "you", really.

sirhelpers.h
johnsonjh: You said left after, so let's go with left.  I like right myself (which is why some of that happened, due to my editor settings).  But I'm not going to die for either :)

sirhelpers.h
johnsonjh: Fixed.

johnsonjh:
Happens to the best of us.  I think I got the fixes you wanted in, let's see.

johnsonjh:
Wewps, signed those commit above with the wrong GPG key, but it's OK.

johnsonjh:
> Thanks for the hard work but this one needs some finesse.

Expected.  Let me know if there's anything else.  I added [Cygwin](https://github.com/aremmell/libsir/pull/65/commits/aa39b9aa5d7f9b9fd3261483aeb29a384b665002) and [Haiku](https://github.com/aremmell/libsir/pull/65/commits/1ee748e5dbde78b5e0337b91281b3cbfc5b69ff0) support on the end (sorry!) but you can easily review those because I squashed them.

sirfilesystem.c
```diff
@@ -218,6 +218,19 @@ char* _sir_getappfilename(void) {
             resolved = false;
             break;
         }
+# elif defined(__HAIKU__)
+        status_t ret = find_path(B_APP_IMAGE_SYMBOL, B_FIND_PATH_IMAGE_PATH, NULL, buffer, PATH_MAX);
```
aremmell, line 222: on a technicality, this should be SIR_MAXPATH, which should be set in sirplatform.h, but if it requires more #ifdefs, then there's no harm. I believe by default it should be set to PATH_MAX outside of windows.

johnsonjh:
I see the issue with scan-build on those three:

> *The reason the analyzer often thinks that a pointer can be null is because the preceding code checked compared it against null. So if you are absolutely sure that it cannot be null, remove the preceding check and, preferably, add an assertion as well.*

Screenshot from 2023-07-06 19-49-02

Reran the analyzer:

> `scan-build: Analysis run complete.`
> `scan-build: Removing directory '/tmp/scan-build-2023-07-06-194818-556339-1' because it contains no reports.`
> `scan-build: No bugs found.`

sirinternal.c
```diff
@@ -1122,6 +1124,10 @@ bool _sir_getthreadname(char name[SIR_MAXPID]) {
         _sir_handleerr(ret);
         return false;
     }
+# if defined(__HAIKU__)
+    if ((strcmp(name, "pthread_func")) || (name[0] == '\0'))
```
aremmell, line 1128: can you please use strncmp(name, "pthread_func", SIR_MAXPID)? and you can use _sir_validstrnofail for the 2nd part; that's what it's for–checks null and first byte.

aremmell (changes requested):
Just a couple minor tweaks, then we're there.

sirfilesystem.c
johnsonjh: It's set in a !__WIN__ block; so using SIR_MAXPATH is fine.

aremmell:
> I see the issue with scan-build on those three:
> 
> > _The reason the analyzer often thinks that a pointer can be null is because the preceding code checked compared it against null. So if you are absolutely sure that it cannot be null, remove the preceding check and, preferably, add an assertion as well._
> 
> Screenshot from 2023-07-06 19-49-02
> 
> Reran the analyzer:
> 
> > `scan-build: Analysis run complete.`
> > `scan-build: Removing directory '/tmp/scan-build-2023-07-06-194818-556339-1' because it contains no reports.`
> > `scan-build: No bugs found.`

Speaking of this, I need your help figuring out what's wrong with my configuration. I think it's special because I'm on macOS and it already provides clang via Xcode.

I found some python package that does what I need and provides scan-build, but it doesn't have scan-view, and doesn't work half the time.

I even tried building clang-tools-extra from source but I can still only get the python one to work.

johnsonjh:
I'll check on mac and see what has to happen there with it..

johnsonjh:
@aremmell Ah, I see on the Mac what maybe works ... 

I installed the `gmake`, `python` (currently 3.11) and `llvm` (currently 16.0.6) packages via Homebrew.  Homebrew installs the Clang bits to a PATH like /usr/local/Cellar/llvm/16.0.6 so it won't conflict with the base system stuff.  (On ARM, I think the PATH is a bit different.)  

I built with `env CC=/usr/local/Cellar/llvm/16.0.6/bin/clang /usr/local/Cellar/llvm/16.0.6/bin/scan-build gmake`

`scan-view` wants to call Python by "python" and not "python3".  I worked around that with: `ln -s /usr/local/bin/python3 $HOME/.local/bin/python`.

Invoke `scan-view` (`/usr/local/Cellar/llvm/16.0.6/bin/scan-view`) with the path, and it does seem to work.

Screenshot from 2023-07-06 20-25-32

johnsonjh:
You can get around that dead assignment via just a (void)x, or the SIR_UNUSED counterpart you have.  I only see that on the Mac, not on Linux.

aremmell:
> @aremmell Ah, I see on the Mac what maybe works ...
> 
> I installed the `gmake`, `python` (currently 3.11) and `llvm` (currently 16.0.6) packages via Homebrew. Homebrew installs the Clang bits to a PATH like /usr/local/Cellar/llvm/16.0.6 so it won't conflict with the base system stuff. (On ARM, I think the PATH is a bit different.)
> 
> I built with `env CC=/usr/local/Cellar/llvm/16.0.6/bin/clang /usr/local/Cellar/llvm/16.0.6/bin/scan-build gmake`
> 
> `scan-view` wants to call Python by "python" and not "python3". I worked around that with: `ln -s /usr/local/bin/python3 $HOME/.local/bin/python`.
> 
> Invoke `scan-view` (`/usr/local/Cellar/llvm/16.0.6/bin/scan-view`) with the path, and it does seem to work.
> 
> Screenshot from 2023-07-06 20-25-32

Makes sense; I didn't try installing llvm through brew, just clang-tools, which doesn't exist. They do have a clang-format package though. Thanks for that; I'll set mine up similarly.

johnsonjh:
Oh, and as you noticed, Clang Analyzer can only lint the active ifdef paths, so it's a good thing you're going to run it on Mac as well as Linux.

When we're done here, I'll lint with PVS and Cppcheck too...

johnsonjh:
> Makes sense; I didn't try installing llvm through brew, just clang-tools, which doesn't exist. They do have a clang-format package though. Thanks for that; I'll set mine up similarly.

Safari sometimes is weird with the Analzyer output... I find Firefox to work better, but it shouldn't matter really.  I did notice that I had some errors on the Mac with Clang-16.  Those were about the macOS logging stuff not being declared right, which I didn't get using the Xcode Clang ... that's worth investigating as well.. probably just a header that you'd have to explicitly include with mainline Clang that is implicitly included with the Apple one.

johnsonjh:
OK, let's see if CI finishes.

PR #66, opened 2023-07-07
Flawfinder

aremmell:
Will you look this over, and make sure I didn't botch anything? The flawfinder stuff is pretty straightforward:

- use strnlen instead of strlen
- use strlcpy/strlcat when available instead of strncpy/strncat

I moved a few things around to better fit the format that was already in place (i.e. where platform-specific includes are placed, etc.), and broke some long lines.

### DIFF
--- a/sirfilesystem.c
+++ b/sirfilesystem.c
@@ -185,7 +185,8 @@ char* _sir_getappfilename(void) {
 #endif
 
 #if !defined(__WIN__)
-# if defined(__linux__) || defined(__NetBSD__) || defined(__SOLARIS__) || defined(__DragonFly__) || defined(__CYGWIN__)
+# if defined(__linux__) || defined(__NetBSD__) || defined(__SOLARIS__) || \
+     defined(__DragonFly__) || defined(__CYGWIN__)
         ssize_t read = readlink(PROC_SELF, buffer, size - 1);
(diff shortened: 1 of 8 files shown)


aremmell:
I would like to assume I can replace `__FreeBSD__` with `__BSD__` in the strl code, but I don't have any other BSDs to test with, so I'll leave that up to you...

johnsonjh:
Sure, I'll test it in the morning, and I'll check on strl availability as well.

johnsonjh:
> I would like to assume I can replace `__FreeBSD__` with `__BSD__` in the strl code, but I don't have any other BSDs to test with, so I'll leave that up to you...

I've see that NetBSD, OpenBSD, and DragonFly BSD all have strlcat and strlcpy.  I'll change the check for \_\_FreeBSD\_\_ to the more general \_\_BSD\_\_.

johnsonjh:
I need to revert the change that includes sys/param.h later.  

That won't work because it breaks the checks for \_\_DragonFly_version and \_\_FreeBSD_version which are defined there, so it needs to happen very early.

johnsonjh (approved):
@aremmell With the two changes I pushed it should be good, pending your OK.

aremmell:
Awesome, thanks for doing that.

PR #72, opened 2023-07-07
Add support for Oracle Developer Studio C and MinGW‑w64.

johnsonjh:
* Tested with [Oracle Developer Studio](https://www.oracle.com/application-development/technologies/developerstudio.html) C/C++ 12.6 on Linux and Solaris.
[]()
[]()
* Tested with [MinGW-w64](https://www.mingw-w64.org/) on Fedora, Ubuntu, and Cygwin, targeting:
  * 64-bit UCRT: `make CC=x86_64-mingw32ucrt-gcc`
  * 64-bit MSVCRT: `make SIR_MSVCRT_MINGW=1 CC=x86_64-mingw32-gcc`
  * 32-bit MSVCRT: `make SIR_MSVCRT_MINGW=1 CC=i686-w64-mingw32-gcc`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,11 +19,42 @@ INSTALLINC = /usr/local/include
 # base CFLAGS
 CFLAGS += -Wall -Wextra -Wpedantic -std=c11 -I. -fPIC
 
+# MinGW compiler-specific flags
+ifneq "$(findstring mingw,$(CC))" ""
+	MINGW ?= 1
+endif
+ifeq ($(MINGW),1)
+  ifneq "$(findstring gcc,$(CC))" ""
+	CFLAGS += -Wno-unknown-pragmas
+  endif
+	MINGW_LIBS=-lshlwapi -lws2_32
+endif
+
+# Oracle compiler-specific flags
+ifneq "$(findstring suncc,$(CC))" ""
+	SUNPRO ?= 1
+endif
+ifeq ($(SUNPRO),1)
+	CFLAGS        += -fcommon
+	FORTIFY_FLAGS ?= -U_FORTIFY_SOURCE
+	MMDOPT        ?= -xMMD
+	PTHOPT        ?= -mt=yes
+else
+	FORTIFY_FLAGS ?= -D_FORTIFY_SOURCE=2
+	MMDOPT        ?= -MMD
+	PTHOPT        ?= -pthread
+endif
+
+# MinGW MSVCRT workaround
+ifeq ($(SIR_MSVCRT_MINGW),1)
+	CFLAGS += -DSIR_MSVCRT_MINGW
+endif
+
 # debug/non-debug CFLAGS
 ifeq ($(SIR_DEBUG),1)
 	CFLAGS += -g -O0 -DDEBUG -U_FORTIFY_SOURCE
 else
-	CFLAGS += -O3 -DNDEBUG -D_FORTIFY_SOURCE=2
+	CFLAGS += -O3 -DNDEBUG $(FORTIFY_FLAGS)
 endif
 
 # enable internal diagnostic logging
@@ -41,11 +72,11 @@ ifeq ($(SIR_NO_SYSTEM_LOGGERS),1)
 endif
 
 # dependencies
-LIBS = -pthread
+LIBS = $(PTHOPT)
 
 # for test rig and example:
 # link with static library, not shared
-LDFLAGS += $(LIBS) -L$(LIBDIR) -lsir_s
+LDFLAGS += $(LIBS) -L$(LIBDIR) -lsir_s $(MINGW_LIBS)
 
 # translation units
 TUS := $(wildcard *.c)
@@ -57,7 +88,7 @@ OBJ  = $(patsubst %, $(INTDIR)/%, $(_OBJ))
 # shared library
 OBJ_SHARED     = $(patsubst %.o, $(INTDIR)/%.o, $(_OBJ))
 OUT_SHARED	   = $(LIBDIR)/libsir.so
-LDFLAGS_SHARED = $(LIBS)
+LDFLAGS_SHARED = $(LIBS) $(MINGW_LIBS)
 
 # static library
 OBJ_STATIC     = $(OBJ_SHARED)
@@ -88,13 +119,13 @@ $(OBJ_TESTS)  : $(OBJ_SHARED)
 $(OBJ_EXAMPLE): $(OBJ_SHARED)
 
 $(OBJ_EXAMPLE): $(EXAMPLE)/$(EXAMPLE).c $(DEPS)
-	$(CC) -MMD -c -o $@ $< $(CFLAGS) -I..
+	$(CC) $(MMDOPT) -c -o $@ $< $(CFLAGS) -I..
 
 $(OBJ_TESTS): $(TESTS)/$(TESTS).c $(DEPS)
-	$(CC) -MMD -c -o $@ $< $(CFLAGS) -I..
+	$(CC) $(MMDOPT) -c -o $@ $< $(CFLAGS) -I..
 
 $(INTDIR)/%.o: %.c $(DEPS)
-	$(CC) -MMD -c -o $@ $< $(CFLAGS)
+	$(CC) $(MMDOPT) -c -o $@ $< $(CFLAGS)
 
 prep:
 	$(shell mkdir -p $(BUILDDIR) && \
--- a/sirconfig.h
+++ b/sirconfig.h
@@ -90,7 +90,11 @@
  *
  * @remark Only applies if ::SIRO_NOPID or ::SIRO_NOTID are not set.
  */
-# define SIR_PIDFORMAT "%d"
+# if defined(__MINGW64__)
+#  define SIR_PIDFORMAT "%lld"
+# else
+#  define SIR_PIDFORMAT "%d"
+# endif
 
 /**
  * The string to place between process and thread IDs.
--- a/sirconfig.h
+++ b/sirconfig.h
@@ -90,11 +90,7 @@
  *
  * @remark Only applies if ::SIRO_NOPID or ::SIRO_NOTID are not set.
  */
-# if defined(__MINGW64__)
-#  define SIR_PIDFORMAT "%lld"
-# else
-#  define SIR_PIDFORMAT "%d"
-# endif
+# define SIR_PIDFORMAT "%d"
 
 /**
  * The string to place between process and thread IDs.
--- a/sirplatform.h
+++ b/sirplatform.h
@@ -88,7 +88,7 @@
 #     define __USE_GNU
 #    endif
 #    if !defined(_GNU_SOURCE)
-#     define _GNU_SOURCE
+#     define _GNU_SOURCE 1
 #    endif
 #    if defined(__clang__) && !defined(_GNU_PTHREAD_H_)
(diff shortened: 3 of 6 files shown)


aremmell (changes requested):
Honestly, if this is going to nuke Makefile, I would rather not take these changes. We could have it in a branch, though.

aremmell:
If the only goal is to require gmake, I'm sure there are easier ways than having two makefiles...

johnsonjh:
It doesn't at nuke anything, that's GitHub showing the diff horribly.

It just renames it to GNUmakefile, so System V and BSD makes won't blow up if you type "make" instead of "gmake".

If you use the wrong "make", you'll get a sane message now:

```
$ make

*** GNU Make is required; try "gmake".

*** Error code 1
The following command caused the error:
false > /dev/null 2>&1
make: Fatal error: Command failed for target `all'
```

Instead of:
```
$ make       
make: "/home/jhj/src/libsir/Makefile" line 23: Invalid line type
make: "/home/jhj/src/libsir/Makefile" line 25: Invalid line type
make: "/home/jhj/src/libsir/Makefile" line 27: Invalid line type
make: "/home/jhj/src/libsir/Makefile" line 30: Invalid line type
make: "/home/jhj/src/libsir/Makefile" line 31: warning: duplicate script for target "ifeq" ignored
make: "Makefile" line 24: warning: using previous script for "ifeq" defined here
make: "/home/jhj/src/libsir/Makefile" line 31: warning: duplicate script for target "(,1)" ignored
make: "Makefile" line 24: warning: using previous script for "(,1)" defined here
make: "/home/jhj/src/libsir/Makefile" line 32: Invalid line type
make: "/home/jhj/src/libsir/Makefile" line 34: Invalid line type
make: "/home/jhj/src/libsir/Makefile" line 35: warning: duplicate script for target "ifeq" ignored
make: "Makefile" line 24: warning: using previous script for "ifeq" defined here
make: "/home/jhj/src/libsir/Makefile" line 35: warning: duplicate script for target "(,1)" ignored
make: "Makefile" line 24: warning: using previous script for "(,1)" defined here
make: "/home/jhj/src/libsir/Makefile" line 36: Invalid line type
make: "/home/jhj/src/libsir/Makefile" line 39: Invalid line type
make: "/home/jhj/src/libsir/Makefile" line 40: warning: duplicate script for target "ifeq" ignored
make: "Makefile" line 24: warning: using previous script for "ifeq" defined here
make: "/home/jhj/src/libsir/Makefile" line 40: warning: duplicate script for target "(,1)" ignored
make: "Makefile" line 24: warning: using previous script for "(,1)" defined here
make: "/home/jhj/src/libsir/Makefile" line 41: Invalid line type
make: Fatal errors encountered -- cannot continue
make: stopped in /home/jhj/src/libsir
```

johnsonjh:
You already require GNU Make, though.  Your use of  "ifeq/else/endif", `$shell`, `$wildcard`, $patsubst`, etc are all GNU-only extensions.

aremmell:
Yeah, I learned that the hard way on FreeBSD, but I'm just saying I would not like to rename it; I've never seen GNUMakefile in my life (I believe you that it works, it's just weird).

Let's see if we can't figure out a way to determine which make executable is running the Makefile and do the same abort procedure?

johnsonjh:
@aremmell Wewps, edited my reply inline instead of replying, sorry.  Let me know what you'd prefer.

aremmell:
> If you have a Makefile that works on all Make variants, it's supposed to be Makefile.

How do you do that if what you said about flow control is true? Maybe there is a way to switch on the make executable? I'll look around on Stack Overflow in a bit.

johnsonjh:
You do that by restricting yourself to POSIX Make, which is a subset of BSD make and GNU make. That's documented at https://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html

aremmell:
This guy says you don't even have to error; you can add this and it'll just redirect: https://stackoverflow.com/a/20308702/1454531

johnsonjh:
> This guy says you don't even have to error; you can add this and it'll just redirect: https://stackoverflow.com/a/20308702/1454531

That makes the assumption that GNU make is "gmake", which often false.

On AIX, it's usually "/opt/freeware/bin/make" (there is common practice but no guarantees that a `gmake` lives in your PATH as a symlink).

On Solaris, you'd use `make -x SUN_MAKE_COMPAT_MODE=GNU`.

The answer there works for FreeBSD though, but only if the user installed the non-default GNU make package.

johnsonjh:
Either way, I don't mind leaving it as is, or having only GNUmakefile (without a fallback compatibility Makefile).

I just think telling someone what they need is GNU Make is less astonishing than trying to automatically run a different command than what the user typed, which that might not even work (and sometimes might not provide any indication as to why.)

Also, you'd fork bomb yourself if you did `gmake -f Makefile` (for some reason).

aremmell:
You're not gonna like this, but this is just pushing me into CMake's welcoming arms.

aremmell:
If it's good enough for LLVM, it's good enough for libsir!

johnsonjh:
Cmake doesn't exist or work on some of the platforms we support, so I'd still have to maintain the makefiles for me. :(

aremmell:
let me guess, some of them don't support autotools either

johnsonjh:
Autotools is somewhat OK though we don't use it.  But the problem is that you're just moving the logic into a shell script and/or M4 rather than in your makefile, and the it generates a POSIX makefile from that.  It's no less work, and it still requires the user to run the script, etc.  IMO it's more trouble than it's worth.

Like on Mac, Homebrew had to patches every single autotools-using package to update it's config.guess and other scripts to work right when Apple moved to ARM, so it's not as portable as you'd think.  CMake is the same thing - you need to get your OS supported in CMake itself or you are out of luck.  

I did some hacking on SerenityOS and they just recently got CMake support (https://gitlab.kitware.com/cmake/cmake/-/commit/45ca894164ce53ada93144131e17ce665267f1ce).  I think that's kind of obnoxious that the tool has all this hard-coded stuff guiding a build.  With make, your makefile itself contains whatever you want for the build so you don't have to modify your build tools to add support for things.

johnsonjh:
> Here's what I think in the end: anyone competent enough with a keyboard to be running BSD or AIX or Solaris or whatever can probably figure out in about 2 milliseconds they need to type gmake instead. I did the first time I saw a million errors come out instead of the normal output.

I'll just remove the fallback then :)  One moment.

johnsonjh:
@aremmell OK, got rid of the fallback.

BSD: `make: no target to make.`
Solaris: `make: Fatal error: No arguments to build`
AIX: `make: make 1254-025 There must be an existing description file or specify a target.`

So that's probably OK.  I can rename it back, but that giant error message is less than helpful.  

And yes, I also assume anyone who uses these is *very* used to invoking the GNU tools and knows when they need to do so.  I know I do.

johnsonjh:
Actually, I'll rename it back and we'll mess with this on another PR. :)

aremmell:
Sorry, I'm back and forth between this and my work laptop.

johnsonjh:
@aremmell OK better.  I *could* add MinGW support to the CI, but to run the tests in Travis, you'd need to use wine (and I don't know how well that's supported), so I didn't.

johnsonjh:
I'm doing nothing else, GitLab HQ is down!

Makefile
```diff
@@ -19,11 +19,42 @@ INSTALLINC = /usr/local/include
...
+	MINGW_LIBS=-lshlwapi -lws2_32
+endif
+
+# Oracle compiler-specific flags
+ifneq "$(findstring suncc,$(CC))" ""
+	SUNPRO ?= 1
+endif
+ifeq ($(SUNPRO),1)
+	CFLAGS        += -fcommon
+	FORTIFY_FLAGS ?= -U_FORTIFY_SOURCE
```
aremmell, line 39: What does `?=` do? I never saw that before. Just `:=`

sirconfig.h
```diff
@@ -90,7 +90,11 @@
  *
  * @remark Only applies if ::SIRO_NOPID or ::SIRO_NOTID are not set.
  */
-# define SIR_PIDFORMAT "%d"
+# if defined(__MINGW64__)
+#  define SIR_PIDFORMAT "%lld"
```
aremmell, line 94: I think it's better to cast to int rather than hacking the macro, tbh. Why did they make PIDs long long? They have 2^64-1 processes?

sirplatform.h
```diff
@@ -145,6 +158,16 @@ int pthread_getname_np(pthread_t thread, char* buffer, size_t length);
 #  include <conio.h>
 #  include <shlwapi.h>
 #  include <direct.h>
+#  if defined(__MINGW32__) || defined(__MINGW64__)
+#   define __USE_MINGW_ANSI_STDIO 1
+typedef  /* Workaround a MinGW bug */
+void (__cdecl* _invalid_parameter_handler)(
+ wchar_t const*, wchar_t const*, wchar_t const*,
+ unsigned int, uintptr_t);
+_invalid_parameter_handler
```
aremmell, line 167: 'sup with dis?

sirplatform.h
```diff
@@ -266,6 +296,10 @@ typedef void (*sir_once_fn)(void);
 
 # else /* __WIN__ */
 
+#  if defined(__MINGW32__) || defined(__MINGW64__)
+#   include <pthread.h>
```
aremmell, line 300: Can you move this up with the other includes for windows or will that break something too?

tests/tests.c
```diff
@@ -1538,7 +1538,8 @@ unsigned sirtest_thread(void* arg) {
 #endif
     }
 
-    printf("\thi, i'm thread id %d, logging to: '%s'...\n", threadid, my_args->log_file);
+    printf("\thi, i'm thread id %lld, logging to: '%s'...\n",
```
aremmell, line 1541: Sorry, I don't wish to do this. It makes no sense to cast on every platform for one that is different; we should cast that to int IMHO.

aremmell (changes requested):

Makefile
johnsonjh: ?= means "set if unset in the environment".  ;)

sirconfig.h
johnsonjh: No idea.  It's a long long int on 64-bit MinGW and just a int on 32-bit MinGW.  The upstream is aware that this is discrepancy compared to native Windows compilers but does not wish to make any changes.  

I see QEMU solves this exact problem with a cast - https://lists.nongnu.org/archive/html/qemu-devel/2015-09/msg06804.html

so I'll do the cast.

sirplatform.h
johnsonjh: MINGW64 provides _invalid_parameter_handler (via UCRT) but does not have a prototype for it (in any header.)

sirplatform.h
johnsonjh: I can move it.

tests/tests.c
johnsonjh: Will cast.

johnsonjh:
@aremmell Hope that's better.  It's all kind of ugly either way.  (But we're almost done!)

johnsonjh:
@aremmell Used a macro and cleaned that up, now it's not horrible.

aremmell (approved):

PR #76, opened 2023-07-08
Suppress Clang Analyzer and Cppcheck warnings

johnsonjh:
@aremmell While it still bothers me a bit still that both analyzers flag this, if you want, we can dismiss with these exclusions.


johnsonjh:
@aremmell It bothers me a bit still that both analyzers flag this, but we can dismiss it with these exclusions.

aremmell (approved):

PR #77, opened 2023-07-08
Add `strlcpy()` and `strlcat()` implementation.

### LINKED ISSUE #74: strncpy warning at sirhelpers.c:206
Enabling GCC LTO and/or PGO can often uncover some additional warnings.

Tested on Linux (Fedora 38) with `gcc (GCC) 13.1.1 20230614 (Red Hat 13.1.1-4)` tested against https://github.com/johnsonjh/libsir-upstream-pr/commit/2e9b5a15dea95ebc77e0947ef5eb1b804721ac1e (#72).

`$ make clean; env CC="gcc" FORTIFY_FLAGS="-D_FORTIFY_SOURCE=3" CFLAGS="-Wformat -Wstringop-truncation -flto=auto -flto-partition=one" LDFLAGS="-flto=auto -flto-partition=one" make`

Results:

```sh
sirhelpers.c: In function ‘__sir_handleerr’:
sirhelpers.c:206:9: warning: ‘__builtin_strncpy’ specified bound 256 equals destination size [-Wstringop-truncation]
  206 |         strncpy(dest, src, count);
      |         ^
```

---

Wonder if you should just use a local strlcpy functions, since you're already using that elsewhere, for systems where it's missing...

johnsonjh:
* Adds OpenBSD reference implementations of `strlcpy()` and `strlcat()` for lacking systems.
  * We will still use the OS provided versions of functions where available, since they might be better optimized or perform additional bounds checking.
* Verified license compatibility; see [FOSSA: ISC License vs. the MIT License](https://fossa.com/blog/open-source-software-licenses-101-isc-license/).
* Inclusion of these `strlcpy()` and `strlcat()` functions do *not* affect the licensing of libsir itself.

Closes #74


.gitignore
```diff
@@ -32,3 +32,6 @@ index.md
 # Visual Studio temporary files
 .vs/
 *.user
+
+# Automatic dependencies
```
aremmell, line 36: These should be generated under `build/`, which is already ignored. Right?

sirstrlcat.h
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
```
aremmell, line 22: header guards?

sirplatform.h
```diff
@@ -164,6 +164,7 @@ int pthread_getname_np(pthread_t thread, char* buffer, size_t length);
 # include <errno.h>
 # include <stdarg.h>
 # include <stdbool.h>
+# include <stddef.h>
```
aremmell, line 197: I don't understand the purpose of this include at this time. Can you explain?

sirstrlcat.h
```diff
@@ -0,0 +1,67 @@
...
+  size_t          n  = dsize;
+  size_t       dlen;
+
+  /* Find the end of dst and adjust bytes left but don't go past end. */
+
+  while (n-- != 0 && *dst != '\0')
+    { dst++; }
+
+  dlen  =   dst - odst;
+  n     = dsize - dlen;
```
aremmell, line 47: This is good, but I will end up reformatting it to better match the style of everything else we have

aremmell (changes requested):
My thoughts on this one:

I think we should create a new unified header for items like this 
(there are sure to be more that crop up in the future, am I right?)

So, instead of sirstrlcat/cpy.h, something like sirimplementation.h and a corresponding source file, so that also fits in with the current way things are done.

Let me know your thoughts.

.gitignore
johnsonjh: I added it because I (somehow) had a `sir.d` appear in my directory.  Could be operator error, but anyway, I just added it by reflex.  We can remove if you want.

sirstrlcat.h
johnsonjh: I'm certainly not married to the formatting in any way. 👍🏻

.gitignore
aremmell: Let's remove it and if you ever see it again we can stick it back in.

sirplatform.h
johnsonjh: The reference code wants it, (apparently) because that's where the [C11 `size_t` definition lives](https://en.cppreference.com/w/c/types).  I guess it's getting pulled in for us via string.h and/or uchar.h and/or wchar.h implicitly, so it might not be strictly required, but since it's C11 I didn't think it would hurt, and I didn't want to have the strl* files include anything themselves.

aremmell: Check out what cppreference.com says about it: https://en.cppreference.com/w/c/types/size_t

Looks like there's a billion ways to get it, and it claims for C11, it's in `uchar.h`...

sirstrlcat.h
johnsonjh: That'll be fine here since sirplatform.h includes them, will add.

johnsonjh:
> My thoughts on this one:
> 
> I think we should create a new unified header for items like this (there are sure to be more that crop up in the future, am I right?)
> 
> So, instead of sirstrlcat/cpy.h, something like sirimplementation.h and a corresponding source file, so that also fits in with the current way things are done.
> 
> Inside the header, we'd have preproc like `SIR_IMPL_STRLCAT` which we would only enable on the correct platforms, and that would compile out the code if it weren't defined.
> 
> Let me know your thoughts.

Could do that, pros and cons to everything .. I'd be careful to *not* build that file if nothing from it is being used, because a lot of compilers will give warnings (or break building with `-Werror` enabled - we don't enable that but some people do), (*i.e.* `file.c: empty translation unit!` on Solaris or `1506-356 (W) Compilation unit is empty.` on AIX), so it would need Makefile changes.  And being static inline'd header-only as it is here ensures no function call overhead (although I'd hope any modern optimizing compiler would be smart enough to automatically inline it.)

So, I'd lean to leave it as-is now, but if we add anything, *then* I'd lean to refactor.

aremmell (changes requested):
Okay, I'll meet you half way:

`sirimpl.h`:
contains the contents of both these headers, and each is guarded with #if defined(SIR_IMPL_FOO).

If warnings can be generated when nothing is defined in the header, I'll figure that out.

sirplatform.h
johnsonjh: Since we aren't using Unicode strings, seems overkill to pull 100+ defintions in ... the entire stddef everywhere is something tiny like:

```c
#ifndef _STDDEF_H
#define _STDDEF_H
#if __cplusplus >= 201103L
#define NULL nullptr
#elif defined(__cplusplus)
#define NULL 0L
#else
#define NULL ((void*)0)
#endif
#define __NEED_ptrdiff_t
#define __NEED_size_t
#define __NEED_wchar_t
#if __STDC_VERSION__ >= 201112L || __cplusplus >= 201103L
#define __NEED_max_align_t
#endif
#include <bits/alltypes.h>
#if __GNUC__ > 3
#define offsetof(type, member) __builtin_offsetof(type, member)
#else
#define offsetof(type, member) ((size_t)( (char *)&(((type *)0)->member) - (char *)0 ))
#endif
#endif
```

I don't mind to try leave it out though, and add it only if any platform gives us a problem.

johnsonjh:
> Okay, I'll meet you half way:
> 
> `sirimpl.h`: contains the contents of both these headers, and each is guarded with #if defined(SIR_IMPL_FOO).
> 
> If warnings can be generated when nothing is defined in the header, I'll figure that out.

Done as header-only it avoids that, since it won't be considered a stand-alone translation unit (it'll be, logically, part of the larger sirplatform.h it gets included from, and that has some unconditional code like typedefs, etc. so it won't give those warnings.  You want to do it, or should I?

.gitignore
johnsonjh: Pushed.

johnsonjh:
> Okay, I'll meet you half way:

Something like the following is better?

johnsonjh:
Try now.

aremmell:
That's precisely what I meant. Thank you for tolerating my OCD, but I have a philosophy about entropy in source code, and I don't change the way I do things or else I find it all goes to shit over time.

aremmell (approved):

aremmell:
Merging when CI finishes.

johnsonjh:
> That's precisely what I meant. Thank you for tolerating my OCD, but I have a philosophy about entropy in source code, and I don't change the way I do things or else I find it all goes to shit over time.

All good!  Thanks for tolerating some Internet rando coming in and molesting all your code!

PR #78, opened 2023-07-08
Don't test os_log activity feature if built with Intel C++ Compiler Classic for macOS

johnsonjh:
ICC doesn't support OS_ACTIVITY_OBJECT_API due to <10.12 target.  (No great loss, as Intel C++ Compiler Classic is deprecated and will be EOL sometime "in the second half of 2023").

### DIFF
--- a/tests/tests.c
+++ b/tests/tests.c
@@ -1137,6 +1137,7 @@ static bool generic_syslog_test(const char* sl_name, const char* identity, const
         pass &= sir_emerg("%d/%d: this emergency message sent to stdout and %s.", i + 1, runs, sl_name);
 
 # if defined(SIR_OS_LOG_ENABLED)
+#  if defined(__APPLE__) && !defined(__INTEL_COMPILER)
         if (i == runs -1 && 0 == strncmp(sl_name, "os_log", 6)) {
             printf("\ttesting os_log activity feature...\n");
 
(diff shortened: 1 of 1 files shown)


aremmell (commented):
> FYI, 10.15 is the oldest supported OS by Apple. Do you actually know anyone using an older version than that?

I don't know anyone still using <10.15, but a few projects are still using this particular compiler, and 10.11 is the latest API level it fully targets.  It also can't do interprocedural optimization (`-ipo`) on macOS >11.  The compiler is fully compatible with macOS Ventura 13.4+ and it works with Xcode versions through 15.0 Beta 2 (though not Beta 3). libsir built with it is fine and and passes all the tests; it just can't be used with the activity grouping bit.

@aremmell  Edit: I think Intel got pretty pissed off at Apple, because when Apple announced the transition to ARM64, they quickly stopped producing compiler updates and are not going to release Intel oneAPI Toolkits or Intel oneAPI DPC++/C++ on macOS.  I guess they aren't under any obligation to produce what is essentially a free/no-cost product to Mac users anymore since Intel dumped them.

aremmell:
10-4. Will merge when Travis completes.

johnsonjh:
Finally finished, hamsters at Travis are busy today.

PR #99, opened 2023-07-10
Use `≳` (equal or equiv. to) in places where we known it's a minimum and greater is OK

johnsonjh:
* Use `≳` (equal or equiv. to) in places where we known it's a minimum and greater is OK


aremmell (approved):
Does >~ mean 'greater than or approximately equal to'?

johnsonjh:
> Does >~ mean 'greater than or approximately equal to'?

Yeah, figured it was the most appropriate

PR #101, opened 2023-07-11
Modify README and regen docs (closes #100)

johnsonjh:
Closes #100 

Please squash when merging

PR #103, opened 2023-07-11
Add support for IBM XL C/C++ V16.1

johnsonjh:
* Add support for IBM XL C/C++ V16.1


johnsonjh:
@aremmell I'll update the compat table after #101 merges or else there would be conflicts

aremmell (approved):

PR #115, opened 2023-07-11
Eliminate possible dangling/empty else condition

johnsonjh:
* Appease Oracle Lint by eliminating possible dangling/empty else condition

### DIFF
--- a/sirerrors.c
+++ b/sirerrors.c
@@ -106,15 +106,17 @@ void __sir_handleerr(int code, const char* func, const char* file, uint32_t line
         /* cppcheck-suppress knownConditionTrueFalse */
         if (0 == finderr && _sir_validstrnofail(message)) {
             __sir_setoserror(code, message, func, file, line);
+#if defined(__HAVE_XSI_STRERROR_R__) || defined(__HAVE_STRERROR_S__)
         } else {
             /* Per my reading of the man pages, GNU strerror_r and normal strerror "can't fail";
              * they simply return a string such as "Unknown nnn error" if unable to look up an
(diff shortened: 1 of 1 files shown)


aremmell:
So, what's the drill? Some compiler doesn't like that if-statement because finderr is only assigned to in certain blocks?

sirerrors.c
```diff
@@ -106,15 +106,17 @@ void __sir_handleerr(int code, const char* func, const char* file, uint32_t line
...
              * error code.
              */
-#if defined(__HAVE_XSI_STRERROR_R__)
+# if defined(__HAVE_XSI_STRERROR_R__)
             _sir_selflog("strerror_r failed! error: %d", finderr);
-#elif defined(__HAVE_STRERROR_S__)
+# elif defined(__HAVE_STRERROR_S__)
             _sir_selflog("strerror_s failed! error: %d", finderr);
+# endif
 #endif
```
aremmell, line 120: Doesn't this #endif need to go outside that bracket?

johnsonjh:
> So, what's the drill? Some compiler doesn't like that if-statement because finderr is only assigned to in certain blocks?

Oracle when being extra picky.

It doesn't like what turns out to be after the ifdef processing:

```
  if (blah) {
    Stuff;
  } else {}
```

sirerrors.c
johnsonjh: image

So if neither defined, turns into:

image

aremmell: Oh, I missed the opener.

PR #117, opened 2023-07-12
Add OpenBSD support

### LINKED ISSUE #109: Add OpenBSD support
* [x] &nbsp; Add OpenBSD support
  * [x] &nbsp; Verify compatibility with system compiler and current Clang and GCC packages
  * [ ] Regenerate docs once merged

johnsonjh:
* Add OpenBSD support.
* Tested on OpenBSD 7.3 with GCC and Clang.

Closes #109


johnsonjh:
Needs cleaning up but is passing tests

johnsonjh:
Let me know if this approach is even right, I know you changed the memory growing stuff for this ... I know you'll want changes.  This seems to work, though.

sirfilesystem.c
```diff
@@ -213,6 +282,27 @@ char* _sir_getappfilename(void) {
...
+        size_t length;
+        int dirname_length;
+        length = _sir_openbsdself(NULL, 0, &dirname_length);
+        if (length < 1) {
+            _sir_handleerr(errno);
+            resolved = false;
+            break;
+        }
+        if (length > size) {
+            resolved = false;
```
aremmell, line 295: You don't need to set resolved = false, just continue when you need to grow.

sirfilesystem.c
```diff
@@ -157,6 +157,75 @@ char* _sir_getcwd(void) {
...
+{
+  char   buffer1[4096];
+  char   buffer2[PATH_MAX];
+  char   buffer3[PATH_MAX];
+  char** argv     = (char**)buffer1;
+  char*  resolved = NULL;
+  int    length   = -1;
+
+  for (;;) {
+    int mib[4] = { CTL_KERN, KERN_PROC_ARGS, getpid(), KERN_PROC_ARGV };
```
aremmell, line 172: So I take it the same mib that we use on FreeBSD doesn't jive on OpenBSD? And it has no readlink?

sirfilesystem.c
```diff
@@ -157,6 +157,75 @@ char* _sir_getcwd(void) {
...
+    size_t size;
+
+    if (sysctl(mib, 4, NULL, &size, NULL, 0) != 0) break;
+    if (size > sizeof(buffer1)) {
+      argv = (char**)malloc(size);
+      if (!argv) break;
+    }
+
+    if (sysctl(mib, 4, argv, &size, NULL, 0) != 0) break;
```
aremmell, line 181: This is the same call as on line 175... shouldn't you only do it again if something changed? I'm assuming the condition: `size > sizeof(buffer1)`?

sirfilesystem.c
```diff
@@ -157,6 +157,75 @@ char* _sir_getcwd(void) {
...
+    if (strchr(argv[0], '/')) {
+      resolved = realpath(argv[0], buffer2);
+      if (!resolved) break;
+    } else {
+      const char* PATH = getenv("PATH");
+      if (!PATH) break;
+      size_t argv0_length = strlen(argv[0]);
+      const char* begin = PATH;
+      while (1) {
```
aremmell, line 190: I would suggest either using `for (;;)` or `while (1)` in both spots to be consistent

sirfilesystem.c
```diff
@@ -157,6 +157,75 @@ char* _sir_getcwd(void) {
...
+      }
+      if (!resolved) break;
+    }
+
+    length = (int)strlen(resolved);
+    if (length <= capacity) {
+      memcpy(out, resolved, (unsigned long)length);
+      if (dirname_length) {
+        int i;
+        for (i = length - 1; i >= 0; --i) {
```
aremmell, line 214: What is this doing? Setting a bunch of the output to '/'? How come?

aremmell (changes requested):
Have some questions about how that function works, and could you please run it through `clang-format` but not any other files? (I had to revert some changes it made and don't wanna have to do that again)

johnsonjh:
Will do.

sirfilesystem.c
johnsonjh: Correct. They don't stash the executed path at all to query later.  And you can't use the readlink trick because they don't have proc.

aremmell:
I don't mean to be a dick, I'm just curious, and always looking for the best way to do things. I found this version on SO, which is like yours kind of, but seems much more simple (thus less prone to errors?) and uses `strtok_r` *(don't be offended-I don't even know if it works)* but if it does, maybe we should take that approach

```c
const char *getExecutablePath(char *epath) {
  int mib[4];
  char **argv;
  size_t len;
  const char *comm;
  int ok = 0;

  mib[0] = CTL_KERN;
  mib[1] = KERN_PROC_ARGS;
  mib[2] = getpid();
  mib[3] = KERN_PROC_ARGV;

  if (sysctl(mib, 4, NULL, &len, NULL, 0) < 0)
    abort();

  if (!(argv = malloc(len)))
    abort();

  if (sysctl(mib, 4, argv, &len, NULL, 0) < 0)
    abort();

  comm = argv[0];

  if (*comm == '/' || *comm == '.') {
    if (realpath(comm, epath))
      ok = 1;
  } else {
    char *sp;
    char *xpath = strdup(getenv("PATH"));
    char *path = strtok_r(xpath, ":", &sp);
    struct stat st;

    if (!xpath)
      abort();

    while (path) {
      snprintf(epath, PATH_MAX, "%s/%s", path, comm);

      if (!stat(epath, &st) && (st.st_mode & S_IXUSR)) {
        ok = 1;
        break;
      }

      path = strtok_r(NULL, ":", &sp);
    }

    free(xpath);
  }

  if (ok)
    *strrchr(epath, '/') = '\0';

  free(argv);
  return ok ? epath : NULL;
}
```

johnsonjh:
> I don't mean to be a dick, I'm just curious, and always looking for the best way to do things. I found this version on SO, which is like yours kind of, but seems much more simple (thus less prone to errors?) and uses `strtok_r` _(don't be offended-I don't even know if it works)_ but if it does, maybe we should take that approach
> 
> ```c
> const char *getExecutablePath(char *epath) {
>   int mib[4];
>   char **argv;
>   size_t len;
>   const char *comm;
>   int ok = 0;
> 
>   mib[0] = CTL_KERN;
>   mib[1] = KERN_PROC_ARGS;
>   mib[2] = getpid();
>   mib[3] = KERN_PROC_ARGV;
> 
>   if (sysctl(mib, 4, NULL, &len, NULL, 0) < 0)
>     abort();
> 
>   if (!(argv = malloc(len)))
>     abort();
> 
>   if (sysctl(mib, 4, argv, &len, NULL, 0) < 0)
>     abort();
> 
>   comm = argv[0];
> 
>   if (*comm == '/' || *comm == '.') {
>     if (realpath(comm, epath))
>       ok = 1;
>   } else {
>     char *sp;
>     char *xpath = strdup(getenv("PATH"));
>     char *path = strtok_r(xpath, ":", &sp);
>     struct stat st;
> 
>     if (!xpath)
>       abort();
> 
>     while (path) {
>       snprintf(epath, PATH_MAX, "%s/%s", path, comm);
> 
>       if (!stat(epath, &st) && (st.st_mode & S_IXUSR)) {
>         ok = 1;
>         break;
>       }
> 
>       path = strtok_r(NULL, ":", &sp);
>     }
> 
>     free(xpath);
>   }
> 
>   if (ok)
>     *strrchr(epath, '/') = '\0';
> 
>   free(argv);
>   return ok ? epath : NULL;
> }
> ```

No problem.  I tried it and it didn't work for quite a few simple cases so I quickly dismissed it.

sirfilesystem.c
johnsonjh: Easy enough.

sirfilesystem.c
johnsonjh: When it iterates through the PATH it sets / if it's missing, so looking for "prog" and given  "/usr/bin:/bin" we check for "/usr/bin/prog" and "/bin/prog" (instead of /binprog").

sirfilesystem.c
johnsonjh: The first time it doesn't actually get the argument value, but queries and sets the size.  It's a brain dead interface IMO.

johnsonjh: IMG_7873

aremmell:
> Let me know if this approach is even right, I know you changed the memory growing stuff for this ... I know you'll want changes. This seems to work, though.

If it passes the test suite, then it definitely works

johnsonjh:
> > Let me know if this approach is even right, I know you changed the memory growing stuff for this ... I know you'll want changes. This seems to work, though.
> 
> If it passes the test suite, then it definitely works

It does :)

> done: all 25 libsir tests passed in 1.301sec!

IMG_7874

johnsonjh:
I was thinking it would be nice to have some more tests, like forcing it to grow the buffer/testing Really Long(tm) filenames, etc.  But you'd probably need to have some kind of way to mock the functions to make that happen, which you wouldn't want in a release library.

aremmell:
> > > Let me know if this approach is even right, I know you changed the memory growing stuff for this ... I know you'll want changes. This seems to work, though.
> > 
> > 
> > If it passes the test suite, then it definitely works
> 
> It does :)
> 
> > done: all 25 libsir tests passed in 1.301sec!
> 
> IMG_7874

After we merge this one, we should Valgrind again after all the changes made this week.

> I was thinking it would be nice to have some more tests, like forcing it to grow the buffer/testing Really Long(tm) filenames, etc. But you'd probably need to have some kind of way to mock the functions to make that happen, which you wouldn't want in a release library.

Before I put the code in libsir, I used that `systest` app I told you about to set the initial buffer size to like 25 and make sure every platform sufficiently grew the buffer and eventually resolved the filename. That's good enough for me...

johnsonjh:
Yes, valgrind is awesome.  I'll also run it through PVS (we have a license for PVS for DPS8M, so to abide by it, you'll have to mention it in any fixes - a small price to pay.)

johnsonjh:
I've been keeping the "worst" for last - AIX.  AIX is the best UNIX, really. I'm a huge fan.  But it's totally different than everything else for this one particular function.

johnsonjh:
We should run Valgrind on illumos, Solaris, FreeBSD, and NetBSD too, not just Linux.

Let me know how you're using it and I can replicate that on the other platforms.

aremmell:
> We should run Valgrind on illumos, Solaris, FreeBSD, macOS, and NetBSD too, not just Linux.
> 
> Let me know how you're using it and I can replicate that on the other platforms.
> 
> Edit: There is also valgrind 3.7.0 for AIX.

Sorry, I was busy failing at creating a basic Xcode project (and I work at Apple). I can't figure it out. If you'd like to try, have at it.

sirfilesystem.c
```diff
@@ -157,6 +157,86 @@ char* _sir_getcwd(void) {
 #endif
 }
 
+#if defined(__OpenBSD__)
```
aremmell, line 160: Just one last nitpick: can you move the implementation to the end of the file and put the declaration at the end of the header with a comment like the others?

aremmell (changes requested):

sirfilesystem.c
johnsonjh: No problem.  Even though it's static?  Or you want it to be made non-static?  I even meant to declare it static inline but forgot.:)

aremmell: oh hmm, well, if you move it to the end of the source file w/o a declaration, getappfilename can't find it right? so i guess make it non-static and declare in the header?

johnsonjh: > oh hmm, well, if you move it to the end of the source file w/o a declaration, getappfilename can't find it right? so i guess make it non-static and declare in the header?

Yep, that works.

aremmell:
Cool, good to merge when Travis finishes in a year.

PR #119, opened 2023-07-13
Fix a typo in the example program

johnsonjh:
* Fix a typo in the example program

### DIFF
--- a/example/example.c
+++ b/example/example.c
@@ -120,7 +120,7 @@ int main(void) {
         "210.10.54.3", 43113, "bob");
 
     sir_warn("Detected downgraded link speed on %s: last transfer rate: %.1f KiB/s",
-        "eth0", "219.4");
+        "eth0", 219.4f);
 
     /* Hmm, what else could go wrong... */
     sir_error("Failed to synchronize with node pool.846.myfooserver.io! Error:"


aremmell (approved):
How did that work?

johnsonjh:
> How did that work?

It worked through the magic of luck (and alignment).  It usually said 0.0 and I never gave it a thought.

And then my little MIPS system said: `[warn] MyFooServer(62211): Detected downgraded link speed on eth0: last transfer rate: 14457935919346138118520009341732972848540587023673880701215377903225352839497160800222978796693163142
383766879915895397872065942685538155642167076290323411809177960448.0 KiB/s` so I decided to look at it. 🤣

johnsonjh:
>  ... so I decided to look at it. 🤣


this makes me think ... perhaps we can leverage the new format verification stuff in GCC.  Pretty sure we can...

johnsonjh:
> >  ... so I decided to look at it. 🤣
> 
> 
> this makes me think ... perhaps we can leverage the new format verification stuff in GCC.  Pretty sure we can... 

Oh yea!  https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-format-function-attribute

aremmell:
It doesn't give me much confidence in `_FORTIFY_SOURCE`, that's for sure. This regressed at some point; I know for sure it worked a few weeks back (or else it would say 0.0 in the sample screenshot).

aremmell:
> > > ... so I decided to look at it. 🤣
> > 
> > 
> > this makes me think ... perhaps we can leverage the new format verification stuff in GCC.  Pretty sure we can...
> 
> Oh yea! https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-format-function-attribute

Sweet. We should create a new issue for this.

johnsonjh:
> It doesn't give me much confidence in `_FORTIFY_SOURCE`, that's for sure. This regressed at some point; I know for sure it worked a few weeks back (or else it would say 0.0 in the sample screenshot).

Fortify is only good for checking bounds pretty much.  The level "3" of fortify is nicer, but not widely supported.  I did check with "3" and nothing bad happened ...

PR #123, opened 2023-07-13
Support compile-time format verification

### LINKED ISSUE #120: Use compiler-assisted format string verification where possible
Since sir takes printf-style strings, some compilers let this fact be annotated, so calls to libsir logging functions will be verified as if they were printf calls, *i.e.* producing helpful warnings on GCC when `-Wformat` is used.

See: https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-format-function-attribute

We could do this like:

```c
#ifdef __GNUC__
# define WFORMAT __attribute__ ((format (printf, 2, 3)))
#else
# define WFORMAT 
#endif
```

... and then put WFORMAT in the magic spots.  We'd extend this for other compatible compilers - they all seem to use something very much like this.  

from the linked docs:
> The parameter archetype determines how the format string is interpreted, and should be printf, scanf, strftime, gnu_printf, gnu_scanf, gnu_strftime or strfmon. (You can also use __printf__, __scanf__, __strftime__ or __strfmon__.) On MinGW targets, ms_printf, ms_scanf, and ms_strftime are also present. archetype values such as printf refer to the formats accepted by the system’s C runtime library, while values prefixed with ‘gnu_’ always refer to the formats accepted by the GNU C Library. On Microsoft Windows targets, values prefixed with ‘ms_’ refer to the formats accepted by the msvcrt.dll library. The parameter string-index specifies which argument is the format string argument (starting from 1), while first-to-check is the number of the first argument to check against the format string. For functions where the arguments are not available to be checked (such as vprintf), specify the third parameter as zero. In this case the compiler only checks the format string for consistency. For strftime formats, the third parameter is required to be zero. Since non-static C++ methods have an implicit this argument, the arguments of such methods should be counted from two, not one, when giving values for string-index and first-to-check. In the example above, the format string (my_format) is the second argument of the function my_print, and the arguments to check start with the third argument, so the correct parameters for the format attribute are 2 and 3.

johnsonjh:
* Support compile-time format verification

Closes #120


johnsonjh:
@aremmell Don't merge yet - but please do see if it a) compiles, and b) actually works on MSVC.

I see conflicting information on if it works with only /ANALYZE or only with Enterprise MSVC or not at all.  I've verified Clang, GCC, and Oracle C.  If there are any C11 supporting versions of these that don't support Wformat, it might also need adjusting, so I have a bit more verification left before merging.

sir.c
```diff
@@ -49,56 +49,64 @@ uint16_t sir_geterror(char message[SIR_MAXERROR]) {
     return _sir_geterrcode(_sir_geterror(message));
 }
 
-bool sir_debug(const char* format, ...) {
+PRINTF_FORMAT_ATTR(1, 2)
```
aremmell, line 52: Do we really need to add it in the header and source file? Seems like it shouldn't be that way... but what do I know; I've never heard of it.

johnsonjh: Yes - declaration in the header and the actual function needs to match for it to both work reliably and not give a warning.

aremmell: Okay, what do the 1 and 2 represent?

johnsonjh:
The "2,0" one might not work on MSVC, not sure if it accepts "0" as an option.  Hopefully we can check.

sir.c
johnsonjh: 1 is the argument that contains the format string specifier and 2 is the first field which contains the arguments.

aremmell:
I know these are macros, and the rule is MACROS_ARE_UCASE, but: my OCD is telling me these are attributes hidden by a macro, and we should name them like attributes: `__attribute_printf_format` would make way more sense to my brain reading it for the first time. What do you think?

johnsonjh:
> I know these are macros, and the rule is MACROS_ARE_UCASE, but: my OCD is telling me these are attributes hidden by a macro, and we should name them like attributes: `__attribute_printf_format` would make way more sense to my brain reading it for the first time. What do you think?

I'm not sure I like it since `__attribute*` might be confused with `__attribute__` or break some parsing, but anything else that is shorter (and lower case) seems weird to me too. Maybe `PRINTF_FORMAT_ATTRIBUTE` and `PRINTF_FORMAT_STRING`?  Those seem too long ...

aremmell:
What is the empty comment for? Is it functional?

aremmell:
Cloning and checking out on Windows...

johnsonjh:
> No compiler warnings w/ MSVC, test rig passes.

If you break a format string, like change the "%zu" back to "%d" or "%u" does it complain and warn?  Do you need to enable "/ANALYZER" (whatever that is?)

at least it doesn't break anything :)

aremmell:
> > No compiler warnings w/ MSVC, test rig passes.
> 
> If you break a format string, like change that "%zu" back to "%d" or "%u" does it complain and warn? Do you need to enable "/ANALYZER" (whatever that is?)
> 
> at least it doesn't break anything :)

`..\tests\tests.c(123,12): warning C4477: 'printf' : format string '%d' requires an argument of type 'int', but variadic argument 1 has type 'size_t'`

johnsonjh:
^ Nice, it works! So that's good - I'm happy with it then.

aremmell (commented):
Is this tested on all your crazy machines? If so, I'll approve it, I just don't want to have to come back and address it again later...

johnsonjh:
> Is this tested on all your crazy machines? If so, I'll approve it, I just don't want to have to come back and address it again later...

Yes, tested all the supported compilers - only exception is I didn't test on GCC 4.8 - won't be able to until tomorrow.  But the documentation days it should work fine on GCC 3.2 and higher, so my confidence level is high.

aremmell (approved):

PR #126, opened 2023-07-13
Leave logs

aremmell:
- Using --leave-logs causes log files to not be deleted after the test suite runs
- Logs are now sent to `./logs/` instead of the working directory
- We need to delete `./logs/*` on clean, and if the `logs` dir isn't present on make tests, we need to mkdir it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ docs/xml/
 index.md
 
 # test artifacts
+*.log
 logs/*
 
 # annoying macOS files
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,7 @@ docs/xml/
 index.md
 
 # test artifacts
-*.log
+logs/*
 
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -24,7 +24,7 @@ docs/xml/
 index.md
 
 # test artifacts
-*.log
```
johnsonjh, line 27: I'd leave *.log in there, just in case, or if ever going back to an older commit.  Can't hurt.

johnsonjh (approved):
Looks good - I'd do #125 first

aremmell:
> Looks good - I'd do #125 first

Yeah, just waiting on Travis, then merging

aremmell:
I am not sure why it failed before–I think the lack of mkdir for logs... I will stop the build on the first failure this time if it fails again.

PR #138, opened 2023-07-14
Remove extra newline

johnsonjh:


aremmell (approved):

aremmell:
Hey you should check out the stuff I just pushed to `plugins`--let me know if you see anything glaringly obvious that we're going to need to modify. I'm assuming `dlopen` and `dlclose` aren't going to work everywhere, but we can start `#if`ing those right now.

What would be super awesome is if you could help with the recursive make I'm going to have to perform. Here's my plan:

- Under `plugins` I'll add a directory `dummy` which contains:
  - dummy_plugin.c
  - dummy_plugin.h
  - Makefile
- We recurse into that (and any other dir under `plugins`) when we `make plugins` and they go to `build/lib` as shared libraries.

I'd like to get that done ASAP because I can't test any of this other code without a dummy plugin. I mean, we don't *need* the Makefile to do that, but it's better than hacking a script together and then throwing it away later.

Thoughts?

aremmell:
> Hey you should check out the stuff I just pushed to `plugins`--let me know if you see anything glaringly obvious that we're going to need to modify. I'm assuming `dlopen` and `dlclose` aren't going to work everywhere, but we can start `#if`ing those right now.
> 
> What would be super awesome is if you could help with the recursive make I'm going to have to perform. Here's my plan:
> 
> * Under `plugins` I'll add a directory `dummy` which contains:
>   
>   * dummy_plugin.c
>   * dummy_plugin.h
>   * Makefile
> * We recurse into that (and any other dir under `plugins`) when we `make plugins` and they go to `build/lib` as shared libraries.
> 
> I'd like to get that done ASAP because I can't test any of this other code without a dummy plugin. I mean, we don't _need_ the Makefile to do that, but it's better than hacking a script together and then throwing it away later.
> 
> Thoughts?

@johnsonjh not sure if you saw this or not.

johnsonjh:
I'll look at it tomorrow (today) and can get the make stuff done for sure. But it's like 5am now I need some sleep first :)

aremmell:
> I'll look at it tomorrow (today) and can get the make stuff done for sure. But it's like 5am now I need some sleep first :)

No worries. I'll create the source files right now in the plugins branch.

PR #148, opened 2023-07-15
Deparallelize on clean and install make targets

johnsonjh:
* Deparallelize on clean and install make targets


johnsonjh:
Will avoid some potential race conditions with make -j N where N>1

aremmell (approved):

PR #156, opened 2023-07-17
Finalize Doxygen docs for external plugin API

aremmell:
Please check for spelling, grammar, punctuation, correctness, formatting, uniformity, etc.

I also made changes to the docs for the file API so that they match.


aremmell:
I've also canceled the CI run for this on purpose. We should do this for any non-code changing PRs, because I'm paying Travis for that service by the usage.

johnsonjh (approved):
👍

PR #157, opened 2023-07-17
Plugin versioning

aremmell:
Just leaving notes for future us, so we remember what we need to do if we make a release, but then decide to add to the plugin interface after the fact.


johnsonjh (approved):
👍

PR #159, opened 2023-07-17
use macros to enter/leave locked sections

aremmell:
This code is used a lot, and it's repetitious. Let's stick to the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and not have the same if-statement everywhere (well, it'll still be there, but at least you can't see it).

### DIFF
--- a/sirfilecache.c
+++ b/sirfilecache.c
@@ -34,17 +34,13 @@ sirfileid _sir_addfile(const char* path, sir_levels levels, sir_options opts) {
     if (!_sir_sanity())
         return NULL;
 
-    sirfcache* sfc = _sir_locksection(SIRMI_FILECACHE);
-    if (!sfc) {
-        _sir_seterror(_SIR_E_INTERNAL);
-        return NULL;
-    }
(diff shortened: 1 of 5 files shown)


sirfilecache.c
```diff
@@ -34,17 +34,13 @@ sirfileid _sir_addfile(const char* path, sir_levels levels, sir_options opts) {
     if (!_sir_sanity())
         return NULL;
 
-    sirfcache* sfc = _sir_locksection(SIRMI_FILECACHE);
-    if (!sfc) {
-        _sir_seterror(_SIR_E_INTERNAL);
-        return NULL;
-    }
+    _SIR_LOCK_SECTION(sirfcache, sfc, SIRMI_FILECACHE, 0);
```
aremmell, line 37: This is 0 instead of NULL in anticipation of #158 being merged.

johnsonjh: +1

johnsonjh (approved):
👍

PR #165, opened 2023-07-18
Coverage tweaks and a small test fixup

johnsonjh:
* Fix a bug in the plugin-loader test.
* Add some coverage exclusions where test coverage is impossible or requires user interactivity.
* Adding missing SPDX headers in `.coveralls.sh`
* Consistently use null comment on empty preprocessor definitions.
* Bump some version numbers for 2.2.1-dev

### DIFF
--- a/src/sirconsole.c
+++ b/src/sirconsole.c
@@ -29,9 +29,9 @@
 #if !defined(__WIN__)
 bool _sir_write_stdio(FILE* stream, const char* message) {
     if (EOF == fputs(message, stream)) {
-        _sir_handleerr(errno); // GCOVR_EXCL_START
+        _sir_handleerr(errno);
         return false;
-    } // GCOVR_EXCL_STOP
+    }
 
     return true;
 }
--- a/src/sirconsole.c
+++ b/src/sirconsole.c
@@ -3,7 +3,7 @@
  *
  * Author:    Ryan M. Lederman <lederman@gmail.com>
  * Copyright: Copyright (c) 2018-2023
- * Version:   2.2.0
+ * Version:   2.2.1
  * License:   The MIT License (MIT)
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy of
--- a/src/sirconsole.c
+++ b/src/sirconsole.c
@@ -29,9 +29,9 @@
 #if !defined(__WIN__)
 bool _sir_write_stdio(FILE* stream, const char* message) {
     if (EOF == fputs(message, stream)) {
-        _sir_handleerr(errno);
+        _sir_handleerr(errno); // GCOVR_EXCL_START
         return false;
-    }
+    } // GCOVR_EXCL_STOP
 
     return true;
 }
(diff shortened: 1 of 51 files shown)


johnsonjh:
@aremmell Not new in this commit, but the test visibly produces the message`[warn]: you won't see this message.`

src/sirconsole.c
```diff
@@ -29,9 +29,9 @@
 #if !defined(__WIN__)
 bool _sir_write_stdio(FILE* stream, const char* message) {
     if (EOF == fputs(message, stream)) {
-        _sir_handleerr(errno);
+        _sir_handleerr(errno); // GCOVR_EXCL_START
```
aremmell, line 32: this may be a legitimate use case, because that code will also (99.9999%) never run

src/sirfilecache.c
```diff
@@ -254,9 +254,9 @@ bool _sirfile_roll(sirfile* sf, char** newpath) {
                         }
 
                         /* if less than one second has elasped since the last roll
-                            * operation, then we'll overwrite the last rolled log file,
-                            * and that = data loss. make sure the target path does not
-                            * already exist. */
```
aremmell, line 259: wth, I *just* fixed this yesterday or the day before

src/sirfilesystem.c
```diff
@@ -182,9 +182,9 @@ char* _sir_getappfilename(void) {
 # endif
     struct stat st;
     if (-1 == lstat(PROC_SELF, &st)) {
-        _sir_handleerr(errno);
+        _sir_handleerr(errno); // GCOVR_EXCL_START
```
aremmell, line 185: We could write a test on these platforms to break this and trigger this code path rather than cheating

src/sirhelpers.c
```diff
@@ -196,11 +196,11 @@ bool _sir_validtextcolor(sir_colormode mode, sir_textcolor color) {
             valid = SIRTC_DEFAULT == color || ((color & 0xff000000) == 0);
             break;
         }
-        case SIRCM_INVALID:
```
aremmell, line 199: This technically should be removed, since you can't set the color mode to this value, and the default is set on startup, but then we'll get linter warnings about not handling that case.

tests/tests.c
```diff
@@ -1730,21 +1730,21 @@ bool sirtest_pluginloader(void) {
...
     printf("\tloading bad plugin: '%s'...\n", plugin6);
-    badid = sir_loadplugin(plugin5);
+    badid = sir_loadplugin(plugin6);
     pass &= 0 == badid;
 
     if (pass)
         print_expected_error();
 
     printf("\tloading bad plugin: '%s'...\n", plugin7);
-    badid = sir_loadplugin(plugin5);
```
aremmell, line 1740: Good catch man

aremmell (changes requested):
Looks good but I want none of those GCOVR comments except in the spot we discussed in filesystem that will never run, and the one I commented on

aremmell (approved):

PR #173, opened 2023-07-20
Increase coverage for builds !SIR_OS_LOG_ENABLED

johnsonjh:
* Increase coverage for builds !SIR_OS_LOG_ENABLED


aremmell (commented):
I will have to spend some time with this; it's not going to just be this straightforward. I'll have documentation/comments I have to update, AND we should probably make sure that other loggers like say systemd don't use categories before we strap it down to os_log. Let's keep this guy on the backburner.

What I would rather do is what I said earlier: in sir/platform.h, the preproc selects SIR_HAS_SYSLOG_CATEGORIES or some such define, which then does what you did here. That way, if we do find another logger to support that has categories, it's just one change to sir/platform.h.

You'll also want to review the code in sirinternal.c related to initializing and updating the system logger. The category is used extensively throughout, and is kind of hard-coded to only support syslog and os_log atm.

johnsonjh:
Yeah, back burner for now is fine.  It's just a whole function that isn't getting tested that could be, so it has almost half a percentage point impact.

aremmell (commented):
Like I was saying–I'll need to dig into this and see what other changes would have to be made, so let's leave it here till I get a chance

PR #174, opened 2023-07-20
Increase coverage of the example program

johnsonjh:
* More coverage for the example program
* Minimal changes and all linters are happy
* Overall a +0.2% gain to 90.195%


example/example.c
```diff
@@ -44,7 +45,13 @@ int report_error(void);
  * @returns EXIT_SUCCESS if execution completes successfully, or EXIT_FAILURE
  * if an error occurs.
  */
-int main(void) {
+int main(int argc, char** argv) {
+#if defined(__HAIKU__) && defined(NDEBUG)
+    disable_debugger(1);
```
aremmell, line 50: what does disaable_debugger do, besides disable the debugger?

example/example.c
```diff
@@ -88,7 +95,7 @@ int main(void) {
      */
     sirfileid fileid = sir_addfile("libsir-example.log", SIRL_ALL, SIRO_NONAME | SIRO_NOHOST);
     if (0 == fileid)
-        report_error();
+        (void)report_error();
```
aremmell, line 98: At this point with this many (void) in front of it, might as well nuke the return type and just not use it as the return value.

example/example.c
```diff
@@ -139,13 +146,13 @@ int main(void) {
 
 #if !defined(SIR_NO_SYSTEM_LOGGERS)
```
aremmell, line 147: Wouldn't this decrease coverage? We want to hit those exports even with NO_SYSTEM_LOGGERS, yeah?

aremmell (dismissed):
tbh I would rather have a lower coverage score than to muddy the example with preprocs and stuff... I dunno, I'll have to see the merged file. Maybe we can put this in another branch for now called say gcovr, and work on coverage there and cherry-pick the stuff we really like into master?

example/example.c
johnsonjh: That's it.  If you don't call that, then by default a graphical debugger interface is attached to any crashing program.  It's nice if you are using the system, but if you connect with SSH or running via a script, the process will just appear to hang forever, no visible crash or core file or anything.

example/example.c
johnsonjh: That's unchanged from the original.  I didn't try to change that (yet).

aremmell: Yeah, we should remove that. Even though we hit those paths during testing, can't hurt and makes the .c look less terrible.

johnsonjh:
> tbh I would rather have a lower coverage score than to muddy the example with preprocs and stuff... I dunno, I'll have to see the merged file. Maybe we can put this in another branch for now called say gcovr, and work on coverage there and cherry-pick the stuff we really like into master?

I'll do something more minimal, tell me what you think.

johnsonjh:
Clearer and doesn't upset the other linters...

johnsonjh:
@aremmell ^

johnsonjh (commented):
Moved other changes to new PR

aremmell (commented):
tbh I'd rather leave the example alone for 0.2%–all I care about is the green :)

johnsonjh:
Well we're only 0.6 or something into the green! :)

johnsonjh:
@aremmell I'll close this out - I have a better idea

PR #175, opened 2023-07-20
Try to unmount proc for a test.

johnsonjh:
* This is a bit brittle; won't work in Docker containers or in chroot-style jails, but hopefully OK in VMs and Travis CI.


johnsonjh:
I don't really like this one because it's a rather brittle test that only gains one or two coverage lines.  Up to you @aremmell

aremmell (approved):

PR #180, opened 2023-07-21
Enable platform-specific (OpenBSD, Solaris, illumos) malloc options in the test suite.

johnsonjh:
* The Solaris and illumos options are documented [here](https://docs.oracle.com/cd/E88353_01/html/E37843/mtmalloc-3malloc.html#REFMAN3Amtmalloc-3malloc).
* The OpenBSD options are documented [here](https://man.openbsd.org/malloc).
* `EXTRA_FLAGS` used since the order matters if building statically and the malloc library should always be last.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ LIBS = $(PTHOPT)
 
 # for test rig and example:
 # link with static library, not shared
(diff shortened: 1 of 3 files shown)


johnsonjh:
@aremmell GTG

aremmell (approved):

PR #181, opened 2023-07-21
Don't crash if `upper_bound` is `0`

johnsonjh:
* The elapsed time is passed to this function, which can be zero because clock_gettime is used to determine the time elapsed. It's possible for clock_gettime to fail, and if that happens it shouldn't cause the test harness to crash.

IMG_7932


johnsonjh:
@aremmell GTG

PR #190, opened 2023-07-23
Several bug fixes, new job queue and thread pool mechanism

aremmell:
* New thread pool.
* Convert remaining unsafe calls (*strlen, etc.*) to safer equivalents. 
* [DUMA](https://github.com/johnsonjh/duma) support for debugging.
* Memory leaks found and fixed.
* Use `CRITICAL_SECTION` as mutex on Windows, and `CONDITION_VARIABLE` instead of hacking it with an event handle.
* Refactor some code for clarity.
* Documentation updates and various spelling corrections. 
* Fix and off-by-one bug in the plugin support code.
* Portability improvements.


johnsonjh (approved):
:) 👍🏻

PR #198, opened 2023-07-29
Improve resolving SIR_MAXHOST value

johnsonjh:
Include `netdb.h` on AIX, Solaris, and illumos. Additionally, include `sys/param.h` on AIX. 
[]()
Allow the user to define SIR_MAXHOST, but if undefined, use:
 1. the larger of MAXHOSTNAMELEN and NI_MAXHOST if both are defined,
 2. MAXHOSTNAMELEN if defined,
 3. NI_MAXHOST if defined,
 4. MAXDNAME if defined and >256,
 5. 256 as a fallback value.


johnsonjh:
Will fix and reopen - not pressing

PR #199, opened 2023-07-29
Don't include `dlfcn.h` if not building plugins

johnsonjh:
Since `dlfcn.h` might not exist (and might be the reason why libsir is being built without plugins); do not include `dlfcn.h` for `SIR_NO_PLUGINS` builds.


aremmell (approved):

johnsonjh:
See #204

PR #201, opened 2023-07-30
Add support for OrangeC (≥6.0.70.93)

johnsonjh:
* Add support for OrangeC (≥6.0.70.93).
* Add a warning for unsupported OrangeC versions.
* Detect if tests are running under Wine, if so,
  * Report the Wine version, and,
  * Adjust some paths accordingly to avoid false positive failures.


johnsonjh:
@aremmell After this is merged, I can add a OrangeC CI job and also enable running the test suite on all the Windows builds via Wine.

johnsonjh:
Will reopen ...

PR #203, opened 2023-07-31
Allow building with D's ImportC

johnsonjh:
Tested working with DMD64 v2.105.0b1 and GDC 13.2.1.


johnsonjh:
Submitting under different PR

PR #204, opened 2023-08-01
Updates for portability and CI 

johnsonjh:
* Allow building with D ImportC
  * Tested working with DMD64 v2.105.0b1 and GDC 13.2.1.
* Add support for Clang-CL
  * Tested with Clang-CL 16.0.6 on Linux using MSVC 17.6 w/Windows 10 SDK headers. The only changes needed are not enabling SAL, and using lowercase (or the exact on-disk capitalization) for library names specified via pragma.
* Add support for OrangeC (≥6.0.70.93)
  * Add a warning for unsupported OrangeC versions.
  * Detect if tests are running under Wine, if so,
    * Report the Wine version, and,
    * Adjust some paths accordingly to avoid false positive failures.
* Don't include `dlfcn.h` if not building plugins
  * Since `dlfcn.h` might not exist (and might be the reason why libsir is being built without plugins); do not include `dlfcn.h` for `SIR_NO_PLUGINS` builds.


tests/tests.c
```diff
@@ -1431,10 +1436,43 @@ bool sirtest_os_log(void) {
...
+#endif
+    return NULL;
+}
+
 bool sirtest_filesystem(void) {
     INIT(si, SIRL_ALL, 0, 0, 0);
     bool pass = si_init;
 
+    /* Wine version */
+    printf("\t_sirtest_wine: '%s'\n", PRN_STR(_sirtest_wine()));
```
aremmell, line 1474: Maybe have this say Running under Wine: %s...", (NULL != winever) ? "yes" : "no"

tests/tests.c
```diff
@@ -1618,8 +1656,8 @@ bool sirtest_filesystem(void) {
 # endif
         {"/dev", true},
 #else /* __WIN__ */
-        {"\\Windows", true},
-        {"\\Program Files", true},
+        {"C:\\Windows", true},
```
aremmell, line 1659: If this isn't needed for Wine or something, \ is translated to "root" or `/` `C:\` (or whatever disk Windows is installed on, which isn't guaranteed to be named C).

aremmell (changes requested):
Sorry, there's a lot to cover here...

include/sir/platform.h
```diff
@@ -26,7 +26,7 @@
 #ifndef _SIR_PLATFORM_H_INCLUDED
 # define _SIR_PLATFORM_H_INCLUDED
 
-# if defined(_MSC_VER)
+# if defined(_MSC_VER) && !defined(__clang__)
```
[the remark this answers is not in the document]
johnsonjh: Purely for building with mainline Clang-CL.  

MSVC always defines `_MSC_VER` but Clang's Clang-CL (https://clang.llvm.org/docs/MSVCCompatibility.html) is differentiated by defining both `__clang__` and `_MSC_VER`. 

Had to add that here so it doesn't try to use the SAL stuff here (only real MSVC supports it).

include/sir.h
```diff
@@ -505,8 +505,9 @@ bool sir_settextstyle(sir_level level, sir_textattr attr, sir_textcolor fg,
  * @see ::sir_settextstyle
  * @see ::default
  *
- * @returns `true` if succcessfully reset, `false` otherwise. Use ::sir_geterror
- *          to obtain information about any error that may have occurred.
+ * @returns bool `true` if succcessfully reset, `false` otherwise. Use
```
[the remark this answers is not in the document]
johnsonjh: Will fix.

src/sirfilesystem.c
```diff
@@ -156,10 +156,20 @@ char* _sir_getcwd(void) {
     return cur;
 # endif
 #else /* __WIN__ */
+# if defined(__ORANGEC__)
+    char cur[SIR_MAXPATH + 1];
```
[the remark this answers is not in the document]
johnsonjh: 1. No, it's using it's own thing for that.  Ugh.
2. It should ...
3. I'll remove the +1, thanks.

include/sir/platform.h
```diff
@@ -282,7 +286,9 @@ _set_thread_local_invalid_parameter_handler(
 # define SIR_MAXHOST 256
 
 # if !defined(__WIN__)
-#  include <dlfcn.h>
+#  if !defined(SIR_NO_PLUGINS)
```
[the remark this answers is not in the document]
johnsonjh: Just this one now, since there were other changes that would have caused conflicts and need rebasing, so I moved it all to this.

tests/tests.c
```diff
@@ -1431,10 +1436,43 @@ bool sirtest_os_log(void) {
 #endif
 }
 
+char *_sirtest_wine(void) {
+#if defined(__WIN__)
+# if defined(_MSC_VER) && !defined(__clang__)
+#  pragma warning(push)
+#  pragma warning(disable : 4152)
+# endif
+    static const char *(CDECL *_p_wine_get_version)(void);
```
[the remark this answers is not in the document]
johnsonjh: Done.

tests/tests.c
```diff
@@ -1431,10 +1436,43 @@ bool sirtest_os_log(void) {
...
+    HMODULE _h_ntdll;
+
+    _h_ntdll = GetModuleHandle("ntdll.dll");
+    if (_h_ntdll != NULL) {
+# if defined(__GNUC__)
+#  pragma GCC diagnostic push
+#  pragma GCC diagnostic ignored "-Wpedantic"
+# endif
+        _p_wine_get_version = (void *)GetProcAddress(_h_ntdll, "wine_get_version");
```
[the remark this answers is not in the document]
johnsonjh: Done.

src/sirfilesystem.c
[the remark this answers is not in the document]
aremmell: Okay, if it can use Win32 API, give me 10 mins and I'll write a version that will work for both

tests/tests.c
```diff
@@ -1431,10 +1436,43 @@ bool sirtest_os_log(void) {
...
+# if defined(__GNUC__)
+#  pragma GCC diagnostic pop
+# endif
+        char *wine_version = (char *)_p_wine_get_version();
+        if (wine_version)
+            return wine_version;
```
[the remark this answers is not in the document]
johnsonjh: Yes, intentional, as wine will always have a "ntdll.dll" loaded (and you can't unload it.)  https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulehandlea says "Retrieves a module handle for the specified module. The module must have been loaded by the calling process.".  So I won't explicitly load or unload it, so, on real Windows, the handle is going to be NULL and the block won't even run, but in the case that ntdll.dll is already mapped, there won't be any "wine_get_version" function to run.

tests/tests.c
johnsonjh: Ah, see, this is failing under Wine, it was always going for the path based on the current working directory, and under wine that is by default "Z:" (`Z:` maps to Unix `/`), so it was looking for `Z:\Windows` and no good.  There is always a `C:\Windows` on Wine.

johnsonjh: @aremmell Not sure how you'd want to handle this one then.

tests/tests.c
```diff
@@ -1431,10 +1436,43 @@ bool sirtest_os_log(void) {
 #endif
 }
 
+char *_sirtest_wine(void) {
```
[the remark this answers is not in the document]
johnsonjh: Done.

tests/tests.c
aremmell: That didn't make sense. Maybe make it say "Wine version: <version> or <not under Wine>"

tests/tests.c
aremmell: I will write a new getcwd. Gimmie 10 mins. If that one still returns Z: then we can leave it C:\

tests/tests.c
johnsonjh: We'll try `printf("\tRunning under Wine: %s\n", sirtest_get_wineversion() ? sirtest_get_wineversion() : "no");`

```
(1/1) 'filesystem'...                                            
                                
        Running under Wine: no
```

Or else it'll show Running under Wine: 8.11 ...

include/sir/platform.h
```diff
@@ -82,6 +82,10 @@
 #    endif
 #   endif
 #  endif
+#  if defined(__IMPORTC__)
```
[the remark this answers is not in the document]
johnsonjh: "ImportC is an implementation of ISO/IEC 9899:2011, which will be referred to as C11; ImportC is a C compiler embedded into the D implementation. It enables direct importation of C files, without needing to manually prepare a D file corresponding to the declarations in the C file. It directly compiles C files into modules that can be linked in with D code to form an executable. It can be used as a C compiler to compile and link 100% C programs."

TL;DR - It's a C11 compiler included with the the Dlang compiler. =)

tests/tests.c
johnsonjh: I'll add OrangeC to the CI/CD job too, so we can see what it does in the test.

tests/tests.c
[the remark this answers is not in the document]
aremmell: Sorry, I misread this as LoadLibrary, not GetModuleHandle

tests/tests.c
```diff
@@ -1431,10 +1436,44 @@ bool sirtest_os_log(void) {
...
+    static const char* (CDECL* _p_wine_get_version)(void);
+    HMODULE _h_ntdll;
+
+    _h_ntdll = GetModuleHandle("ntdll.dll");
+    if (_h_ntdll != NULL) {
+# if defined(__GNUC__)
+#  pragma GCC diagnostic push
+#  pragma GCC diagnostic ignored "-Wpedantic"
+# endif
+        _p_wine_get_version = (void *)GetProcAddress(_h_ntdll, "wine_get_version");
```
aremmell, line 1454: Is this the line that triggers warning 4152? Check out the example here. I believe if you do it like this, you won't get a warning (use STDCALL instead of CDECL, and typedef that function signature, then cast to that type not void*)

https://learn.microsoft.com/en-us/cpp/build/getprocaddress?view=msvc-170

tests/tests.c
```diff
@@ -1431,10 +1436,44 @@ bool sirtest_os_log(void) {
...
+# if defined(__GNUC__)
+#  pragma GCC diagnostic pop
+# endif
+        char *wine_version = (char *)_p_wine_get_version();
```
aremmell, line 1458: Plz asterisks to left of space 😬

aremmell (changes requested):

tests/tests.c
johnsonjh: I'll check it out shortly, but it needs to work on more than just MSVC ... so pretty sure the warning still happens elsewhere, let me see ...

include/sir/platform.h
```diff
@@ -446,6 +451,24 @@ typedef BOOL(CALLBACK* sir_once_fn)(PINIT_ONCE, PVOID, PVOID*);
...
 # endif
 
+# if defined(__WIN__) && defined(__ORANGEC__)
+#  if defined(__ORANGEC_MAJOR__) && defined(__ORANGEC_MINOR__) && defined(__ORANGEC_PATCHLEVEL__)
+#   if __ORANGEC_MAJOR__ <= 6
+#    if __ORANGEC_MINOR__ <= 70
+#     if __ORANGEC_PATCHLEVEL__ <= 92
+#      if !defined(ORANGEC_VERSION_WARNING)
+#       warning OrangeC versions before 6.0.70.93 are unsupported.
+#       define ORANGEC_VERSION_WARNING
```
aremmell, line 461: Shouldn't we just #error here? what's the point of warning and continuing?

johnsonjh: Earlier versions can be made to work, but they're missing some functionality like strnlen - up to you. :)

tests/tests.c
johnsonjh: @aremmell I'm stuck downstairs for a bit away from a real keyboard so if you want to try it and we'll see what CI thinks, feel free, unless I get to it first

aremmell (approved):

PR #208, opened 2023-08-02
Run Windows tests using Wine for CI; use betterC mode for DMD/ImportC build.

johnsonjh:
* Run Windows tests using Wine for CI.
* Use betterC for DMD/ImportC build.
* Improve ImportC compatibility with older versions of DMD.


include/sir/platform.h
```diff
@@ -84,7 +84,24 @@
 #  endif
 #  if defined(__IMPORTC__)
 #   undef __HAVE_ATOMIC_H__
+#   undef __SIZEOF_INT128__
 #   define _BITS_FLOATN_H 1
+#   undef __builtin_bswap16
+#   define __builtin_bswap16
+#   undef __builtin_bswap32
+#   define __builtin_bswap32
+#   undef __builtin_bswap64
```
aremmell, line 93: I think it's about time we split `platform.h` up just like we did with `Makefile`... it's getting unruly.

I would like to move all these platform-specific preprocessor blocks into their own `name-platform.h` that is `#include`d in the main header.

tests/tests.c
```diff
@@ -1678,12 +1678,10 @@ bool sirtest_filesystem(void) {
     }
 
     /* checking file descriptors. */
-    static const int bad_fds[] = {
-        0,
```
aremmell, line 1682: clang-format? I like my arrays 1 on a line :)

johnsonjh: Fixed

include/sir/platform.h
johnsonjh: It's not too bad right now, at least all the ugliness is kept in one place.  And since the ordering matters, having to split things up wouldn't necessarily make it more readable, I don't think.  There are five different sections that switch on _AIX, for example.

aremmell: Sooner or later (probably sooner rather than later), it's gonna get out of hand... if necessary we can have `name-platform-n.h` and include them in the appropriate places. Readability isn't my concern–it's maintainability of that gnarly header file with 8 million preprocessor switches

tests/tests.c
```diff
@@ -1678,12 +1678,15 @@ bool sirtest_filesystem(void) {
...
 
     /* checking file descriptors. */
-    static const int bad_fds[] = {
+    static int bad_fds[] = {
         0,
         1,
         2,
         1234
     };
+    if (sirtest_get_wineversion()) { //-V547
```
aremmell, line 1687: What's 547?

include/sir/platform.h
aremmell: We can merge this for now, but I'm not super stoked on adding like 100 lines of preproc for a compiler nobody I know has or ever will use :P

aremmell (approved):

tests/tests.c
johnsonjh: Always false, since it will never be true (on Unix).

PR #209, opened 2023-08-05
Add support for GNU/Hurd and older glibc versions

johnsonjh:
* Support for older glibc versions (45c4d2bdb1dd1cc7ed4682bac4f4cf3bfc3d6789)
  * Verified with Debian 5, glibc 2.7 (2007)
  * Should work with glibc as old as 2.4 (2006)
* Support for GNU/Hurd (5b51a2c7e4897f72e219fc97c60b9ce8513db032)
  * Tested on Debian GNU/Hurd 2023 and Debian GNU/Hurd Unstable
  * Working with GCC (9.5–13.2) and Clang (9.0.1–13.0.1)
* Document that the Portable C Compiler is supported (2ca40aa2c577ecd6fed431a35a1ff35ac76dfc13)
  * Tested with PCC 1.2.0.DEVEL‑20230730 on Linux
  * Probably working on BSD systems also, will update once confirmed


aremmell (changes requested):
gtg except change message for inability to get thread names

aremmell (approved):

PR #213, opened 2023-08-09
Support Embarcadero C++ for Windows; add to CI

### LINKED ISSUE #215: Add Embarcadero C++ job to CI/CD
Options are ...
[]()
* Embarcadero C++ 7.60 (RAD Studio / C++Builder 11.3 Alexandria)
* Embarcadero C++ 7.20 (C++ Compiler 10.2 Tokyo)

johnsonjh:
* Tested with:
  * Embarcadero C++ 7.60 (C++Builder 11.3 Alexandria)
  * Embarcadero C++ 7.20 (C++ Compiler 10.2 Tokyo)

Closes #215

### DIFF
--- a/src/sirfilesystem.c
+++ b/src/sirfilesystem.c
@@ -27,7 +27,13 @@
 #include "sir/internal.h"
 
 #if defined(__WIN__)
-# pragma comment(lib, "shlwapi.lib")
+# if (defined(__TURBOC__) || defined(__BORLANDC__) || \
+     defined(__BCPLUSPLUS__) || defined(__CODEGEARC__)) && \
+     defined(_WIN64)
+#  pragma comment(lib, "shlwapi.a")
+# else
+#  pragma comment(lib, "shlwapi.lib")
+# endif
 #endif
 
 bool _sir_pathgetstat(const char* restrict path, struct stat* restrict st, sir_rel_to rel_to) {
--- a/src/sirfilesystem.c
+++ b/src/sirfilesystem.c
@@ -27,7 +27,7 @@
 #include "sir/internal.h"
 
 #if defined(__WIN__)
-# if defined(__EMBARCADEROC___) && defined(_WIN64)
+# if defined(__EMBARCADEROC__) && defined(_WIN64)
 #  pragma comment(lib, "shlwapi.a")
 # else
 #  pragma comment(lib, "shlwapi.lib")
--- a/src/sirfilesystem.c
+++ b/src/sirfilesystem.c
@@ -27,7 +27,13 @@
 #include "sir/internal.h"
 
 #if defined(__WIN__)
-# pragma comment(lib, "shlwapi.lib")
+# if (defined(__TURBOC__) || defined(__BORLANDC__) || \
+     defined(__BCPLUSPLUS__) || defined(__CODEGEARC__)) && \
+     defined(_WIN64)
+#  pragma comment(lib, "shlwapi.a")
+# else
+#  pragma comment(lib, "shlwapi.lib")
+# endif
 #endif
 
 bool _sir_pathgetstat(const char* restrict path, struct stat* restrict st, sir_rel_to rel_to) {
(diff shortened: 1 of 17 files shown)


johnsonjh:
In case you need it for testing:
* [Embarcadero C++ 7.20 / C++ Compiler 10.2 Tokyo](https://www.embarcadero.com/free-tools/ccompiler) - Free Edition
* [Embarcadero C++ 7.60 / RAD Studio/C++Builder 11.3 Alexandria](https://docwiki.embarcadero.com/RADStudio/Alexandria/en/Community_Edition) - Community Edition

johnsonjh:
@aremmell https://github.com/aremmell/libsir/pull/213/commits/f3f05f8663e126f4f13147e516fa366ae71619e6 is a *way* nicer hack than mine, good job.

src/sirhelpers.c
```diff
@@ -341,7 +341,11 @@ bool _sir_getchar(char* input) {
         return false;
 
 #if defined(__WIN__)
+# if defined(__EMBARCADEROC__)
```
aremmell, line 344: What's up with this? It can't call _getch?

src/sirinternal.c
```diff
@@ -33,7 +33,11 @@
 #include "sir/mutex.h"
 
 #if defined(__WIN__)
-# pragma comment(lib, "ws2_32.lib")
```
aremmell, line 36: This is making me want to pull all of these comment(lib) pragmas into one file... say, sirplatform.c...

tests/tests.c
```diff
@@ -1291,10 +1291,10 @@ static bool generic_syslog_test(const char* sl_name, const char* identity, const
     for (int i = 0; i < runs; i++) {
         /* randomly skip setting process name, identity/category to thoroughly
          * test fallback routines; randomly update the config mid-run. */
-        bool set_procname = getrand_bool((uint32_t)sirtimerelapsed(&timer));
-        bool set_identity = getrand_bool((uint32_t)sirtimerelapsed(&timer));
-        bool set_category = getrand_bool((uint32_t)sirtimerelapsed(&timer));
-        bool do_update    = getrand_bool((uint32_t)sirtimerelapsed(&timer));
+        bool set_procname = getrand_bool(1+(uint32_t)sirtimerelapsed(&timer));
```
aremmell, line 1294: why is 1+ needed here? is your machine too fast to get random results without it or something?

tests/tests.c
```diff
@@ -1629,6 +1622,12 @@ bool sirtest_filesystem(void) {
             printf("\t" GREEN("_sir_ispathrelative('%s') = %s") "\n", abs_or_rel_paths[n].path,
                 relative ? "true" : "false");
         }
+
+        pass &= ret;
+        if (!ret) {
```
aremmell, line 1627: Why move this guy?

tests/tests.c
```diff
@@ -2207,11 +2209,17 @@ bool filter_error(bool pass, uint16_t err) {
 }
 
 uint32_t getrand(uint32_t upper_bound) {
-#if !defined(__WIN__)
+#if !defined(__WIN__) || defined(__EMBARCADEROC__)
 # if defined(__MACOS__) || defined(__BSD__)
+    if (upper_bound < 1)
```
aremmell, line 2214: upper_bound could only be 0 if it were < 1? are you seeing zeroes?

aremmell (changes requested):

src/sirhelpers.c
johnsonjh: Nope, only getch()!

tests/tests.c
johnsonjh: It passes that to the random function, which takes an upper bound.  And the number returned  is usually upper_bound - 1.  So I want to be sure it's >1.

The man page for arc4random_uniform says:

> An upper_bound of 0 doesn't make sense in a call to arc4random_uniform(). Such a call will fail, *and* return 0. Be careful, since that value is not less than upper_bound. In some cases, such as accessing an array, using that value could result in Undefined Behavior.

So I'm just being cautious.

tests/tests.c
johnsonjh: The arc4random page says to never ever pass a 0 so I'm just ensuring we never do.  As an aside, that function is available in newer versions of Linux and OpenBSD - should we look into using it when it's available?  I guess
Random() is fine here.

tests/tests.c
aremmell: Well, sirtimerelapsed is returning milliseconds since the start, so in this case, there is no way it will ever be < 10-100 let alone 1. There may be other spots where we use getrand_bool that are more susceptible to that problem though...

tests/tests.c
aremmell: But it also says that it returns a value less than upper bound, and uint32_t can't hold decimal places... so shouldn't it be a 2?

aremmell: And yes, we should use it wherever it's available... I'll check my lunixes. I think it's already enabled for FreeBSD

aremmell: This should be a 2, not a 1, if it returns "less than upper_bound"

tests/tests.c
aremmell: And with the adjustment you made in getrand, we're good and don't need a +1 here; this value will always be at least >10. you can log it if you don't believe

aremmell: I'm not gonna budge on this one ;)

tests/tests.c
aremmell: Not gonna let this one go either. You said that you should never pass a zero, and the man page says it returns less than upper_bound, which means if you pass 1 it can only return 0. At least with 2 it could be 0 or 1

src/sirfilesystem.c
```diff
@@ -82,16 +86,30 @@ bool _sir_pathgetstat(const char* restrict path, struct stat* restrict st, sir_r
...
+# endif
+
         stat_ret = stat(abs_path, st);
         _sir_safefree(&base_path);
     } else {
         stat_ret = stat(path, st);
     }
 #endif
-    if (-1 == stat_ret && ENOENT == errno)
+    if (-1 == stat_ret && (ENOENT == errno || ENOTDIR == errno))
```
aremmell, line 109: Good call. I didn't see that error in the man page.

aremmell (changes requested):
Just remove the `1+` and change the 1 to a 2 and it looks good

johnsonjh (commented):
Done.

aremmell (approved):

PR #221, opened 2023-08-19
Avoid undefined behavior and integer overflow

johnsonjh:
* Avoid undefined behavior and integer overflow
* Add new CI jobs ...
  * Clang Memory Sanitizer
  * Clang Undefined Behavior Sanitizer (32-bit)
  * Clang Undefined Behavior Sanitizer (64-bit)
  * Clang Address Sanitizer (32-bit)
  * Clang Address Sanitizer (64-bit)
  * MSVC i686 (32-bit)

### DIFF
--- a/include/sir/helpers.h
+++ b/include/sir/helpers.h
@@ -325,7 +325,7 @@ bool _sir_getchar(char* input);
  */
 static inline
 uint32_t FNV32_1a(const uint8_t* data, size_t len) {
-    uint32_t hash = 2166136261UL;
+    uint32_t hash = 2166136261U;
     for (size_t n = 0; n < len; n++) {
         hash ^= (uint32_t)data[n];
         hash *= 16777619UL;
--- a/include/sir/helpers.h
+++ b/include/sir/helpers.h
@@ -337,6 +337,9 @@ uint32_t FNV32_1a(const uint8_t* data, size_t len) {
  * Implementation of the 64-bit FNV-1a OWHF (http://isthe.com/chongo/tech/comp/fnv/)
  * watered down to only handle null-terminated strings.
  */
+# if defined(__clang__) /* only Clang has unsigned-integer-overflow */
+SANITIZE_SUPPRESS("unsigned-integer-overflow")
+# endif
 static inline
 uint64_t FNV64_1a(const char* str)
 {
--- a/include/sir/helpers.h
+++ b/include/sir/helpers.h
@@ -328,7 +328,7 @@ uint32_t FNV32_1a(const uint8_t* data, size_t len) {
     uint32_t hash = 2166136261U;
     for (size_t n = 0; n < len; n++) {
         hash ^= (uint32_t)data[n];
-        hash *= 16777619UL;
+        hash = (uint32_t)(hash * 16777619ULL);
     }
     return hash;
 }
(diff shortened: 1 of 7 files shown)


johnsonjh:
@aremmell Marked as a draft until I add the Sanitizer checks to CI.

include/sir/helpers.h
```diff
@@ -328,7 +328,7 @@ uint32_t FNV32_1a(const uint8_t* data, size_t len) {
     uint32_t hash = 2166136261U;
     for (size_t n = 0; n < len; n++) {
         hash ^= (uint32_t)data[n];
-        hash *= 16777619U;
+        hash = (uint32_t)(hash * 16777619ULL);
```
aremmell, line 331: I'm not sure I understand the utility in promoting the constant to 64-bit if we're to fit it in 32 bits. Shouldn't it be just `U`, or does adding ULL promote `hash` to unsigned long long as well?

include/sir/platform.h
```diff
@@ -53,6 +53,16 @@
 #  endif
 # endif
 
+# if defined(__has_attribute) && (defined(__clang__) || defined(__GNUC__))
+#  if __has_attribute(no_sanitize)
+#   define SANITIZE_SUPPRESS(str) __attribute__((no_sanitize(str)))
+#  else
```
aremmell, line 59: Seems like this could be simplified;

```c
#if (stuff)
# if (otherstuff)
#   define thing(str) fancy(str)
# endif
#endif

#if !defined(thing)
# define thing(str)
#endif
```

src/sirhelpers.c
```diff
@@ -93,6 +93,9 @@ bool _sir_validfd(int fd) {
 }
 
 /** Validates a sir_update_config_data structure. */
+# if defined(__clang__) /* only Clang has implicit-conversion; GCC BZ#87454 */
+SANITIZE_SUPPRESS("implicit-conversion")
```
aremmell, line 97: I will need to know what the implicit conversion warnings in every place we are putting this are

tests/tests.c
```diff
@@ -926,7 +929,7 @@ bool sirtest_optionssanity(void) {
     }
 
     /* greater than SIRO_NOHDR. */
-    invalid = (0xFFFF0000 & ~SIRO_NOHDR);
+    invalid = (0xFFFF0000 & ~SIRO_NOHDR); /* implicit-conversion */
```
aremmell, line 932: `sir_options` = uint32_t, `SIRO_NOHDR` = int?

aremmell (commented):
Just need to see the implicit conversion warnings for these places where we are sticking fugly preprocessor stuff, so I can try to fix them rather than doing that, and simplify the preproc in platform.h, otherwise looks good

johnsonjh:
Just comment them out and rebuild locally with the sanitizer variables / command line from the .gitlab-ci.yml.

Any implicit conversion warning isn't necessarily a sign of a problem - it's potentially something to look at.

include/sir/helpers.h
johnsonjh: I'd have to check exactly what it was, but without it, it fails the sanitizer test due to an overflow.

aremmell (approved):

PR #225, opened 2023-08-21
Portability enhancements

johnsonjh:
* Improve Makefile portability and argument ordering
* Adjust integer constant syntax for compatibility with C++11 with user-defined literals
  * See https://en.cppreference.com/w/cpp/language/user_literal#Notes
* Support building with G++
  * Tested with G++ 13.2.1 on Linux/x86_64
* Support building with Clang++
  * Tested with Clang++ 16.0.6 on Linux/x86_64
* Support building with Circle
  * Tested with Circle/C++ Build 200 on Linux/x86_64
  * Both C and C++ modes are supported
* Support building with Oracle Studio C++ 12.6
  * Tested with Oracle Studio C++ 12.6 on Linux/x86_64


example/example.c
```diff
@@ -182,6 +182,6 @@ int main(void) {
 int report_error(void) {
     char message[SIR_MAXERROR] = {0};
     uint16_t code              = sir_geterror(message);
-    fprintf(stderr, "libsir error: (%"PRIu16", %s)\n", code, message);
+    fprintf(stderr, "libsir error: (%" PRIu16", %s)\n", code, message);
```
aremmell, line 185: Don't you need a space on both sides if you need spaces?

src/sirerrors.c
```diff
@@ -175,7 +175,7 @@ uint32_t _sir_geterror(char message[SIR_MAXERROR]) {
         char* heap_msg = NULL;
 
         if (_SIR_E_PLATFORM == sir_errors[_mid].e) {
-            heap_msg = calloc(SIR_MAXERROR, sizeof(char));
+            heap_msg = (char *)calloc(SIR_MAXERROR, sizeof(char));
```
aremmell, line 178: No spaces before *

src/sirhelpers.c
```diff
@@ -355,14 +355,14 @@ bool _sir_getchar(char* input) {
 # endif
      return true;
 #else /* !__WIN__ */
-    struct termios cur = {0}, new = {0};
+    struct termios cur = {0}, xnew = {0};
```
aremmell, line 358: I presume this is because new is a keyword. Let's use tnew

src/sirinternal.c
```diff
@@ -246,23 +250,23 @@ bool _sir_init_sanity(const sirinit* si) {
 }
 
 static
-bool _sir_updatelevels(const char* name, sir_levels* old, sir_levels* new) {
-    if (*old != *new) {
-        _sir_selflog("updating %s levels from %04"PRIx16" to %04"PRIx16, name, *old, *new);
-        *old = *new;
+bool _sir_updatelevels(const char* name, sir_levels* old, sir_levels* xnew) {
```
aremmell, line 253: Let's use lnew

src/sirinternal.c
```diff
@@ -246,23 +250,23 @@ bool _sir_init_sanity(const sirinit* si) {
...
+        *old = *xnew;
     } else {
-        _sir_selflog("skipped superfluous update of %s levels: %04"PRIx16, name, *old);
+        _sir_selflog("skipped superfluous update of %s levels: %04" PRIx16, name, *old);
     }
     return true;
 }
 
 static
-bool _sir_updateopts(const char* name, sir_options* old, sir_options* new) {
```
aremmell, line 260: onew

tests/tests.c
```diff
@@ -720,7 +720,7 @@ bool sirtest_textstylesanity(void) {
     bool pass = si_init;
 
     printf("\t" WHITEB("--- explicitly invalid ---") "\n");
-    pass &= !sir_settextstyle(SIRL_INFO, 0xbbb, 800, 920);
+    pass &= !sir_settextstyle(SIRL_INFO, (sir_textattr)0xbbb, 800, 920);
```
aremmell, line 723: Why does sir_textattr need a cast? And shouldn't it be TA_CAST

aremmell (dismissed):
If this is all just to support compiling with a C++ compiler I don't want to take it into master; we can create a branch that supports C++ compilers but this isn't C++, and nobody uses circle :)

johnsonjh:
Almost all of it is just whitespace related due to the difference in how C++11 preprocessors handle string combining vs. what is valid in just C.

Makefile changes are mostly improving efficiency and saner ordering.

I'm not adding any C++ specific stuff at all (besides the right extern C thing so plug-in symbols aren't mangled).

The Circle-specific stuff is in platform_circle.h.  The other changes are casts that are made explicit that were implicit - and somehow it makes it benchmark about 10% faster here too (with Clang and GCC).

I'll go over your things individually though, it's not as terrible as you'd imagine.

example/example.c
johnsonjh: It only cares about the leading spaces, oddly.  Cppreference explains the dumb rationale somewhere, I think at where I linked it, if you care to know.  Hah.

src/sirerrors.c
johnsonjh: Can adjust

src/sirhelpers.c
johnsonjh: Yeah.  It's an error in C++ but it's also a warning in MISRA-C and some other pedantic style enforcements.

Those style guides are interesting though, they don't like variables that are bare adjectives (new, old, good, bad); they want them to be self-describing like old_something, good_thing, etc.

But yeah, I'll make this tnew.

tests/tests.c
johnsonjh: Yes to the latter, I'll fix that.

But why? Seems this is some Circle weirdness.  But since it's not wrong, I'll play along.  

Here's the error it throws without it:

```c
error: tests/tests.c:723
cannot convert prvalue int to sir_textattr
    pass &= !sir_settextstyle(SIRL_INFO, 0xbbb, 800, 920);
                                         ^
```

I actually really like some of the warnings and diagnostics that the Circle parser reports.  It doesn't have near as many warnings / diagnostics as Clang or GCC sadly.

PR #230, opened 2023-08-23
Use localtime_r for improved performance

johnsonjh:
* Use `localtime_r` for improved performance

### DIFF
--- a/include/sir/config.h
+++ b/include/sir/config.h
@@ -123,7 +123,7 @@
  *   15:13:41 Fri 9 Jun 23 (-0600)
  *   ~~~
  */
-# define SIR_FHTIMEFORMAT "%H:%M:%S %a %d %b %y (%z)"
+# define SIR_FHTIMEFORMAT "%H:%M:%S %a %d %b %Y %Z (%z)"
 
 /**
  * The format string written to a log file when logging begins or the file
--- a/include/sir/config.h
+++ b/include/sir/config.h
@@ -123,7 +123,7 @@
  *   15:13:41 Fri 9 Jun 23 (-0600)
  *   ~~~
  */
-# define SIR_FHTIMEFORMAT "%H:%M:%S %a %d %b %Y %Z (%z)"
+# define SIR_FHTIMEFORMAT "%H:%M:%S %a %d %b %y (%z)"
 
 /**
  * The format string written to a log file when logging begins or the file
--- a/include/sir/platform.h
+++ b/include/sir/platform.h
@@ -202,6 +202,7 @@ int pthread_getname_np(pthread_t thread, char* buffer, size_t length);
 #  endif
 # else /* _WIN32 */
 #  define __WIN__
+#  undef SIR_NO_SYSTEM_LOGGERS
 #  define SIR_NO_SYSTEM_LOGGERS
 #  undef __HAVE_ATOMIC_H__
 #  define __WANT_STDC_SECURE_LIB__ 1
--- a/src/sirhelpers.c
+++ b/src/sirhelpers.c
@@ -323,7 +323,7 @@ struct tm* _sir_localtime(const time_t* restrict timer, struct tm* restrict buf)
         }
 # else
         struct tm* ret = localtime_s(timer, buf);
-        if (0 == ret) {
+        if (!ret) {
             (void)_sir_handleerr(ret);
             return NULL;
         }
--- a/src/sirhelpers.c
+++ b/src/sirhelpers.c
@@ -327,10 +327,10 @@ struct tm* _sir_localtime(const time_t* restrict timer, struct tm* restrict buf)
 
         return buf;
 #else /* !__HAVE_STDC_SECURE_OR_EXT1___ */
-# if defined(__WIN__) && !defined(__EMBARCADEROC__)
-        struct tm* ret = localtime(timer);
-# else
+# if !defined(__WIN__) || defined(__EMBARCADEROC__)
         struct tm* ret = localtime_r(timer, buf);
+# else /* __WIN__ */
+        struct tm* ret = localtime(timer);
 # endif
         if (!ret)
             (void)_sir_handleerr(errno);
--- a/src/sirhelpers.c
+++ b/src/sirhelpers.c
@@ -315,15 +315,15 @@ int _sir_fopen(FILE* restrict* restrict streamptr, const char* restrict filename
 struct tm* _sir_localtime(const time_t* restrict timer, struct tm* restrict buf) {
     if (_sir_validptr(timer) && _sir_validptr(buf)) {
 #if defined(__HAVE_STDC_SECURE_OR_EXT1__) && !defined(__EMBARCADEROC__)
-# if defined(__WIN__)
-        errno_t ret = localtime_s(buf, timer);
-        if (0 != ret) {
-            (void)_sir_handleerr(ret);
-            return NULL;
-        }
-# else
+# if !defined(__WIN__)
         struct tm* ret = localtime_s(timer, buf);
         if (!ret) {
+            (void)_sir_handleerr(errno);
+            return NULL;
+        }
+# else /* __WIN__ */
+        errno_t ret = localtime_s(buf, timer);
+        if (0 != ret) {
             (void)_sir_handleerr(ret);
             return NULL;
         }
--- a/src/sirhelpers.c
+++ b/src/sirhelpers.c
@@ -314,25 +314,24 @@ int _sir_fopen(FILE* restrict* restrict streamptr, const char* restrict filename
 
 struct tm* _sir_localtime(const time_t* restrict timer, struct tm* restrict buf) {
     if (_sir_validptr(timer) && _sir_validptr(buf)) {
-#if defined(__HAVE_STDC_SECURE_OR_EXT1__)
+#if defined(__HAVE_STDC_SECURE_OR_EXT1___)
 # if defined(__WIN__)
         errno_t ret = (errno_t)localtime_s(buf, timer);
+# else
+        errno_t ret = (errno_t)localtime_s(timer, buf);
+# endif
         if (0 != ret) {
             (void)_sir_handleerr(ret);
             return NULL;
         }
 
         return buf;
-# else /* __WIN__ */
-        struct tm* ret = localtime_s(timer, buf);
-        if (!ret)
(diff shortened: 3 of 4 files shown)


include/sir/config.h
```diff
@@ -123,7 +123,7 @@
  *   15:13:41 Fri 9 Jun 23 (-0600)
```
aremmell, line 123: Comment is out of sync. I don't really think it's necessary to change the format either; people generally know it's not 1923 or 2023, and that their time zone is -600 or whatever.

src/sirhelpers.c
```diff
@@ -323,16 +323,15 @@ struct tm* _sir_localtime(const time_t* restrict timer, struct tm* restrict buf)
         }
 
         return buf;
-# else /* __WIN__ */
-        struct tm* ret = localtime_s(timer, buf);
+# else /* !__WIN__ */
```
aremmell, line 326: Could you flip it around? We always do if !__WIN__ first because that's the primary target...

aremmell (changes requested):

johnsonjh:
Before:
Screenshot from 2023-08-23 09-24-56

After:
Screenshot from 2023-08-23 09-24-51

40.9068% decrease in time for _sir_format with this change:
Screenshot from 2023-08-23 09-29-46

Measurable improvement:
Screenshot from 2023-08-23 09-32-32

src/sirhelpers.c
```diff
@@ -314,25 +314,24 @@ int _sir_fopen(FILE* restrict* restrict streamptr, const char* restrict filename
 
 struct tm* _sir_localtime(const time_t* restrict timer, struct tm* restrict buf) {
     if (_sir_validptr(timer) && _sir_validptr(buf)) {
-#if defined(__HAVE_STDC_SECURE_OR_EXT1__)
+#if defined(__HAVE_STDC_SECURE_OR_EXT1___)
```
aremmell, line 317: 2 underscores at the end, not 3, unless it is defined that way elsewhere in which case it's wrong there too

src/sirhelpers.c
```diff
@@ -314,25 +314,24 @@ int _sir_fopen(FILE* restrict* restrict streamptr, const char* restrict filename
...
+# else
+        errno_t ret = (errno_t)localtime_s(timer, buf);
+# endif
         if (0 != ret) {
             (void)_sir_handleerr(ret);
             return NULL;
         }
 
         return buf;
+#else /* !__HAVE_STDC_SECURE_OR_EXT1___ */
```
aremmell, line 329: ditto here

src/sirhelpers.c
```diff
@@ -314,25 +314,24 @@ int _sir_fopen(FILE* restrict* restrict streamptr, const char* restrict filename
 
 struct tm* _sir_localtime(const time_t* restrict timer, struct tm* restrict buf) {
     if (_sir_validptr(timer) && _sir_validptr(buf)) {
-#if defined(__HAVE_STDC_SECURE_OR_EXT1__)
+#if defined(__HAVE_STDC_SECURE_OR_EXT1___)
 # if defined(__WIN__)
         errno_t ret = (errno_t)localtime_s(buf, timer);
```
aremmell, line 319: This shouldn't need a cast to `errno_t` on windows; that is the type it returns... and according to [cppreferfence](https://en.cppreference.com/w/c/chrono/localtime), on other platforms it should return `struct tm*` and the return value is reverse in meaning. NULL (0) means it failed, whereas [on windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-s-localtime32-s-localtime64-s?view=msvc-170), 0 means success. Thanks MSFT.

src/sirhelpers.c
```diff
@@ -314,25 +314,24 @@ int _sir_fopen(FILE* restrict* restrict streamptr, const char* restrict filename
...
             (void)_sir_handleerr(ret);
             return NULL;
         }
 
         return buf;
+#else /* !__HAVE_STDC_SECURE_OR_EXT1___ */
+# if !defined(__WIN__) || defined(__EMBARCADEROC__)
+        struct tm* ret = localtime_r(timer, buf);
 # else /* __WIN__ */
-        struct tm* ret = localtime_s(timer, buf);
```
aremmell, line 327: Yeah this is wrong, and shouldn't even have compiled on windows... see the link to the docs in above comment

aremmell (changes requested):

src/sirhelpers.c
johnsonjh: Fixed

src/sirhelpers.c
johnsonjh: Fixed

src/sirhelpers.c
johnsonjh: On my cell but I'll fix - I'm sure I'll mess it up the first time, :)

include/sir/platform.h
```diff
@@ -202,6 +202,7 @@ int pthread_getname_np(pthread_t thread, char* buffer, size_t length);
 #  endif
 # else /* _WIN32 */
 #  define __WIN__
+#  undef SIR_NO_SYSTEM_LOGGERS
```
aremmell, line 205: This reminds me... when I was running the test rig earlier, SIR_OS_LOG_ENABLED was not defined, so I think something broke at some point. I will take a look.

src/sirhelpers.c
```diff
@@ -314,22 +314,26 @@ int _sir_fopen(FILE* restrict* restrict streamptr, const char* restrict filename
...
-        errno_t ret = (errno_t)localtime_s(timer, buf);
-# endif
+        errno_t ret = localtime_s(buf, timer);
         if (0 != ret) {
             (void)_sir_handleerr(ret);
             return NULL;
         }
+# else
+        struct tm* ret = localtime_s(timer, buf);
+        if (0 == ret) {
```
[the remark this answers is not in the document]
johnsonjh: Fixed

include/sir/platform.h
aremmell: We should definitely add it to the CI pipeline, if unable to do so in Docker, I can set up a VM on my iMacPro and we can figure out how to run it there

src/sirhelpers.c
```diff
@@ -314,25 +314,28 @@ int _sir_fopen(FILE* restrict* restrict streamptr, const char* restrict filename
...
-        return buf;
-# else /* __WIN__ */
+# else
         struct tm* ret = localtime_s(timer, buf);
-        if (!ret)
-            (void)_sir_handleerr(errno);
-
-        return ret;
+        if (!ret) {
+            (void)_sir_handleerr(ret);
```
aremmell, line 327: I don't think you meant to do this–`ret` is `struct tm*`. My man page says:

```
When encountering an error, these
     functions return NULL and set errno to an appropriate value
```

So it was right before...

src/sirhelpers.c
```diff
@@ -314,25 +314,28 @@ int _sir_fopen(FILE* restrict* restrict streamptr, const char* restrict filename
 
 struct tm* _sir_localtime(const time_t* restrict timer, struct tm* restrict buf) {
     if (_sir_validptr(timer) && _sir_validptr(buf)) {
-#if defined(__HAVE_STDC_SECURE_OR_EXT1__)
+#if defined(__HAVE_STDC_SECURE_OR_EXT1__) && !defined(__EMBARCADEROC__)
 # if defined(__WIN__)
```
aremmell, line 318: & I hate to be a pain in the ass, but can  you switch the order of the preproc blocks? We consistently do if !defined(__WIN__) first, since the majority of the time that will be the branch it takes

aremmell (changes requested):

src/sirhelpers.c
johnsonjh: Looking at https://en.cppreference.com/w/c/chrono/localtime for the localtime_s function, it says it returns "a copy of the buf pointer, or null pointer on error (which may be a runtime constraint violation or a failure to convert the specified time to local calendar time)".  So checking for `!ret` would be correct, no? I should be using errno though in the error reporting here, fixed.

src/sirhelpers.c
johnsonjh: No prob

src/sirhelpers.c
johnsonjh: Kind of a pain, I don't have any EXT1 UNIXen here to actually test that path... I think maybe VxWorks?  Grr.

src/sirhelpers.c
```diff
@@ -314,25 +314,28 @@ int _sir_fopen(FILE* restrict* restrict streamptr, const char* restrict filename
 
 struct tm* _sir_localtime(const time_t* restrict timer, struct tm* restrict buf) {
     if (_sir_validptr(timer) && _sir_validptr(buf)) {
-#if defined(__HAVE_STDC_SECURE_OR_EXT1__)
-# if defined(__WIN__)
-        errno_t ret = (errno_t)localtime_s(buf, timer);
+#if defined(__HAVE_STDC_SECURE_OR_EXT1__) && !defined(__EMBARCADEROC__)
```
aremmell, line 317: I think it would be better to put the embarcadero preproc where have stdc secure is defined in the first place, then we don't have to do these checks in the source

src/sirhelpers.c
```diff
@@ -314,25 +314,28 @@ int _sir_fopen(FILE* restrict* restrict streamptr, const char* restrict filename
...
-        if (!ret)
-            (void)_sir_handleerr(errno);
-
-        return ret;
-# endif
-#else
-        SIR_UNUSED(buf);
+#else /* !__HAVE_STDC_SECURE_OR_EXT1__ */
+# if !defined(__WIN__) || defined(__EMBARCADEROC__)
+        struct tm* ret = localtime_r(timer, buf);
```
aremmell, line 335: Windows only has a _s huh and no _r?

johnsonjh: Correct, they have _s and have no _r (and their _s is thread-safe).

src/sirhelpers.c
johnsonjh: It's done that way because their libc should have __HAVE_STDC_SECURE_OR_EXT1__ defined, except the localtime_s they provide doesn't work like the windows one, but they do provide a localtime_r, which Microsoft does not, and it's faster, like on Unix.

PR #232, opened 2023-08-23
Use (and pass) Clang's ` -Wextra-semi-stmt` check and enable in CI

johnsonjh:
* Use (and pass) Clang's `-Wextra-semi-stmt` check and enable in CI


aremmell (approved):
That missing semicolon on the one BIN_SEARCH macros is going to keep me up at night.

PR #234, opened 2023-08-23
Use (and pass) both Clang's `-Wformat-nonliteral` and GCC's `-Wformat-nonliteral -Wformat-truncation` and enable in CI

johnsonjh:
* Use (and pass) Clang's `-Wformat-nonliteral` and enable in CI
* Use (and pass) GCC's `-Wformat-nonliteral -Wformat-truncation` and enable in CI

### DIFF
--- a/.lint.sh
+++ b/.lint.sh
@@ -205,6 +205,7 @@ test_extra()
                     -Wdouble-promotion
                     -Wmissing-variable-declarations
                     -Wconversion
+                    -Wextra-semi-stmt
                     -Wno-sign-conversion
                     -Wswitch-enum
                     -Wno-string-conversion
--- a/.lint.sh
+++ b/.lint.sh
@@ -205,6 +205,7 @@ test_extra()
                     -Wdouble-promotion
                     -Wmissing-variable-declarations
                     -Wconversion
+                    -Wbad-function-cast
                     -Wextra-semi-stmt
                     -Wno-sign-conversion
                     -Wswitch-enum
--- a/.lint.sh
+++ b/.lint.sh
@@ -236,7 +236,7 @@ test_gccextra()
       rm -f ./.extra.sh
       env CC="${CCACHE:-env} gcc" \
         "${MAKE:-make}" \
-        -j "${CPUS:-1}" \
+        -j 1 \
         mcmb
       printf '%s' 'true' > ./.extra.sh
       # shellcheck disable=SC2090,SC2086,SC2016
@@ -254,7 +254,7 @@ test_gccextra()
                     -Wformat-nonliteral
                     -Wformat-truncation"
             ${MAKE:-make}
-                -j "${CPUS:-1}" ' | tr '\n' ' ' | tr -s ' ' >> ./.extra.sh
+                -j 1 ' | tr '\n' ' ' | tr -s ' ' >> ./.extra.sh
       printf '%s\n' ' && true' >> ./.extra.sh
       chmod a+x ./.extra.sh
       sh ./.extra.sh
--- a/.lint.sh
+++ b/.lint.sh
@@ -203,6 +203,7 @@ test_extra()
             CFLAGS="-Werror
                     -Wmissing-prototypes
                     -Wdouble-promotion
+                    -Wmissing-variable-declarations
                     -Wconversion
                     -Wno-sign-conversion
                     -Wswitch-enum
--- a/.lint.sh
+++ b/.lint.sh
@@ -227,9 +227,9 @@ test_gccextra()
     || {
       printf '%s\n' \
         "NOTICE: gcc is required for the gccextra-warning check."
-      NO_EXTRAWARN=1
+      NO_EXTRAGCCWARN=1
     }
-  test -z "${NO_EXTRAWARN:-}" \
+  test -z "${NO_EXTRAGCCWARN:-}" \
     && {
       printf '%s\n' "building with extra-warning flags ..."
       env "${MAKE:-make}" clean
--- a/.lint.sh
+++ b/.lint.sh
@@ -224,7 +224,7 @@ test_extra()
 test_gccextra()
 {
   test_mcmb
-  command -v gcc > /dev/null 2>&1 \
+  gcc --version 2>&1 | grep -qi GCC > /dev/null 2>&1 \
     || {
       printf '%s\n' \
         "NOTICE: gcc is required for the gccextra-warning check."
--- a/.lint.sh
+++ b/.lint.sh
@@ -207,7 +207,52 @@ test_extra()
                     -Wconversion
                     -Wno-sign-conversion
                     -Wswitch-enum
-                    -Wno-string-conversion"
+                    -Wno-string-conversion
+                    -Wformat-nonliteral"
+            ${MAKE:-make}
+                -j "${CPUS:-1}" ' | tr '\n' ' ' | tr -s ' ' >> ./.extra.sh
+      printf '%s\n' ' && true' >> ./.extra.sh
+      chmod a+x ./.extra.sh
+      sh ./.extra.sh
+      rm -f ./.extra.sh
+    }
+}
+
+################################################################################
+
+test_gccextra()
+{
+  test_mcmb
+  command -v gcc > /dev/null 2>&1 \
+    || {
+      printf '%s\n' \
+        "NOTICE: gcc is required for the gccextra-warning check."
+      NO_EXTRAWARN=1
+    }
+  test -z "${NO_EXTRAWARN:-}" \
+    && {
+      printf '%s\n' "building with extra-warning flags ..."
+      env "${MAKE:-make}" clean
+      rm -f ./.extra.sh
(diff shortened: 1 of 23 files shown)


johnsonjh:
This one is interesting - enable "`-Wformat-nonliteral`" for Clang, and "`-Wformat-nonliteral -Wformat-truncation`" for GCC.

As it is now, master, with Clang:
```sh
src/sirerrors.c:180:56: warning: format string is not a string literal [-Wformat-nonliteral]
                (void)snprintf(heap_msg, SIR_MAXERROR, sir_errors[_mid].msg, sir_te.os_error,
                                                       ^~~~~~~~~~~~~~~~~~~~
/usr/include/bits/stdio2.h:61:31: note: expanded from macro 'snprintf'
                            __glibc_objsize (str), __VA_ARGS__)
                                                   ^~~~~~~~~~~
```
... and GCC:
```sh
src/sirerrors.c: In function ‘_sir_geterror’:
src/sirerrors.c:181:21: warning: format not a string literal, argument types not checked [-Wformat-nonliteral]
  181 |                     (_sir_validstrnofail(sir_te.os_errmsg) ? sir_te.os_errmsg : SIR_UNKNOWN));
      |                     ^
src/sirinternal.c: In function ‘_sir_formattime’:
src/sirinternal.c:1066:9: warning: format not a string literal, format string not checked [-Wformat-nonliteral]
 1066 |         _sir_localtime(&now, &timebuf));
      |         ^~~~~~~~~~~~~~
```

So we will handle the first one, and define a `_SIR_E_PLATFORM_ERRORFORMAT` to makes Clang happy.  This doesn't handle GCC, though.
```
include/sir/errors.h:91:38: warning: ‘%s’ directive output may be truncated writing up to 255 bytes into a region of size between 223 and 233 [-Wformat-truncation=]
   91 | # define _SIR_E_PLATFORM_ERRORFORMAT "Platform error code %d: %s"
      |                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/sirerrors.c:180:56: note: in expansion of macro ‘_SIR_E_PLATFORM_ERRORFORMAT’
  180 |                 (void)snprintf(heap_msg, SIR_MAXERROR, _SIR_E_PLATFORM_ERRORFORMAT, sir_te.os_error,
      |                                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~
include/sir/errors.h:91:63: note: format string is defined here
   91 | # define _SIR_E_PLATFORM_ERRORFORMAT "Platform error code %d: %s"
      |                                                               ^~
In file included from /usr/include/stdio.h:906,
                 from include/sir/platform.h:278,
                 from include/sir/types.h:29,
                 from include/sir/helpers.h:29,
                 from include/sir/errors.h:29:
In function ‘snprintf’,
    inlined from ‘_sir_geterror’ at src/sirerrors.c:180:23:
/usr/include/bits/stdio2.h:54:10: note: ‘__builtin___snprintf_chk’ output between 24 and 289 bytes into a destination of size 256
   54 |   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   55 |                                    __glibc_objsize (__s), __fmt,
      |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   56 |                                    __va_arg_pack ());
      |                                    ~~~~~~~~~~~~~~~~~
```

In this case, since truncation here will likely never happen and is also just fine, we can suppress this. The [GCC authors advise the following macro be used](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89312#c5) when truncation is expected:
```c
#define snprintf_trunc(dst, size, ...) \
  do { \
    volatile size_t n = size; \
    snprintf (dst, n, __VA_ARGS__); \
  } while (0)
```
I've added a variation of this to `include/sir/helpers.h`.

For the second one, it seems that Clang is either smart enough to know that the format strftime attribute we set means that the third argument passed to strftime will be a format string, or it doesn't do the extra check that GCC does.  For this one, the only way I know to silence this is to use a suppression, which I added.  Maybe you can think of a better way?

Anyway, this makes both Clang and GCC happy with the extra checks, which I have enabled in CI.

include/sir/errors.h
```diff
@@ -114,7 +116,7 @@ static const struct {
     {_SIR_E_PLUGINDAT, "Data produced by plugin is invalid"},
     {_SIR_E_PLUGINVER, "Plugin interface version unsupported"},
     {_SIR_E_PLUGINERR, "Plugin reported failure"},
-    {_SIR_E_PLATFORM,  "Platform error code %d: %s"},
+    {_SIR_E_PLATFORM,  _SIR_E_PLATFORM_ERRORFORMAT},
```
aremmell, line 119: That actually squashes the warning? What if we change the entry in the map to 'const char* const'?

include/sir/helpers.h
```diff
@@ -333,6 +333,15 @@ uint32_t FNV32_1a(const uint8_t* data, size_t len) {
     return hash;
 }
 
+/**
+ * Wrapper for snprintf when truncation is intended.
+ */
+# define snprintf_truncate(dst, size, ...) \
+    do { \
+      volatile size_t n = size; \
+      (void)snprintf (dst, n, __VA_ARGS__); \
```
aremmell, line 342: No space before parens please. And why do you have to make a volatile copy of size?

src/sirinternal.c
```diff
@@ -1062,8 +1062,17 @@ bool _sir_formattime(time_t now, char* buffer, const char* format) {
         return _sir_seterror(_SIR_E_INVALID);
 
     struct tm timebuf = {0};
+#if defined(__GNUC__) && !defined(__clang__) && \
+    !(defined(__OPEN64__) || defined(__OPENCC__))
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wformat-nonliteral"
+#endif
     size_t fmttime    = strftime(buffer, SIR_MAXTIME, format,
```
aremmell, line 1070: Same thing here. What if format is const char* const? Does that alleviate it?

aremmell (changes requested):
Have questions

include/sir/helpers.h
johnsonjh: The GCC authors provided this macro as the preferred way to silence this warning when truncation is expected or desired.  There is a link at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89312 that goes more into it.  There isn't yet any alternative I've seen that works in all cases.

include/sir/errors.h
johnsonjh: I just tried, no difference.

src/sirinternal.c
johnsonjh: No, same

include/sir/helpers.h
```diff
@@ -335,6 +335,15 @@ uint32_t FNV32_1a(const uint8_t* data, size_t len) {
...
```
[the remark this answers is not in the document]
aremmell: Probably rename to `_sir_snprintf_trunc` as well, just to keep it homogenous.

src/sirinternal.c
aremmell, line 1070: only 1 space between fmttime and = please

aremmell (changes requested):
Some minor tweaks

.lint.sh
```diff
@@ -220,6 +221,50 @@ test_extra()
 
 ################################################################################
 
+test_gccextra()
+{
+  test_mcmb
+  command -v gcc > /dev/null 2>&1 \
```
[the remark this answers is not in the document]
johnsonjh: K

include/sir/errors.h
```diff
@@ -88,6 +88,8 @@ enum sir_errorcode {
 # define _SIR_E_PLATFORM  _sir_mkerror(SIR_E_PLATFORM)
 # define _SIR_E_UNKNOWN   _sir_mkerror(SIR_E_UNKNOWN)
 
+# define _SIR_E_PLATFORM_ERRORFORMAT "Platform error code %d: %s"
+
```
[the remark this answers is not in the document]
johnsonjh: Some are used all over ... via `return _sir_seterror(_SIR_E_...)`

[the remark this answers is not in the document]
johnsonjh: IMG_8164

.lint.sh
```diff
@@ -220,6 +221,50 @@ test_extra()
 
 ################################################################################
 
+test_gccextra()
+{
+  test_mcmb
+  gcc --version 2>&1 | grep -qi GCC > /dev/null 2>&1 \
+    || {
+      printf '%s\n' \
+        "NOTICE: gcc is required for the gccextra-warning check."
```
aremmell, line 230: Did you want a space between gcc and extra-warning?

aremmell (approved):

PR #237, opened 2023-08-24
Fix SIR_OS_LOG_ENABLED

aremmell:
- Detection was broken at the preprocessor level. Fixed it so that if the target system is macOS and the compiler is Clang, define.
- Fixed two implicit conversion warnings that we didn't catch because this preprocessor macro was never enabled.


johnsonjh (approved):
That'll work!

PR #241, opened 2023-08-29
Change sir_level and sir_option to defines, rather than enum

### LINKED ISSUE #223: Change sir_level[s] and sir_option[s] to typedefs and #defines
The ubsan warnings we are seeing are due to the fact that these are enums, which have an implicit type of `int`.

Remove the enums (they are not scoped anyhow, so it'll be not much work), and replace with typedef uint32_t, and #defines that have a `U` suffix

aremmell:
If it works, closes #223

### DIFF
--- a/include/sir/helpers.h
+++ b/include/sir/helpers.h
@@ -328,7 +328,7 @@ bool _sir_getchar(char* input);
 static inline
 uint32_t FNV32_1a(const uint8_t* data, size_t len) {
     uint32_t hash = 2166136261U;
-    for (size_t n = 0ul; n < len; n++) {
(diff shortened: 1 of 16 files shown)


johnsonjh:
Looks like it needs some explicit casts - fails CI due to integer overflow - https://gitlab.com/libsir/libsir/-/pipelines/984744443/failures

johnsonjh (approved):
👍🏻

I usually use caps (`1UL`, `1L`, etc.) so things like `1l` won't look like `11` or `ll` but otherwise all looks OK.

johnsonjh (approved):
Looking good, tested locally, will merge.

PR #243, opened 2023-08-29
Performance improvement via attribute hinting

johnsonjh:
* Before: 411172.9 lines/sec
* After: 425647.2 lines/sec
* +3.52% improvement

*Average of 60 perf runs (30 each, writing to tmpfs)*


johnsonjh:
Tested again, still same performance difference with GCC (~+3.6%)

aremmell (commented):
I would love to say I had the same experience as you, but on macOS w/ AppleClang, I am not getting a perf increase–I might even be getting a decrease.

I'm guessing you used GCC? I'll try Clang on Fedora.

johnsonjh:
Closing - more benchmarking shows it's not worth it to get a small gain for GCC at the expense of Clang, and the last commit with a simpler patch only gets 0.3% or so gain with no change on Clang on two Intel systems.

PR #245, opened 2023-08-29
Change more types to uintxx_t instead of enum

aremmell:
- SIRU_
- SIRSL_
- Also pass `0U` to `INIT` in tests to avoid warnings


johnsonjh (approved):
👍🏻

PR #250, opened 2023-08-29
>= 1u

aremmell:
Maybe this isn't always true. If it is, I will just change them back to signed integers.


johnsonjh (approved):
Seems fine by me - we'll see what Coverity thinks tomorrow I guess, if you merge it.

PR #252, opened 2023-08-30
Get rid of unneeded case statements

aremmell:
`SIRL_*` are not an enum anymore, so all members don't need to be present in a switch.


johnsonjh (approved):
👍🏻

PR #261, opened 2023-09-07
Coverity: Fix for unchecked return values

johnsonjh:
* Coverity: Fix for unchecked return values


tests/tests.c
```diff
@@ -454,7 +454,7 @@ bool sirtest_filecachesanity(void) {
     if (pass)
         print_expected_error();
 
-    sir_info("test test test");
+    (void)sir_info("test test test");
```
aremmell, line 457: this should be a _sir_eqland(pass, sir_info())

aremmell (approved):

aremmell (approved):

tests/tests.c
johnsonjh: Ah, you got it there before I saw this, thanks.

PR #264, opened 2023-09-07
Return false if text styles fail to reset

johnsonjh:
* Return false if text styles fail to reset


src/sirinternal.c
```diff
@@ -141,10 +141,7 @@ bool _sir_init(sirinit* si) {
     _sir_initialize_stdio();
 #endif
 
-    _sir_setcolormode(SIRCM_16);
-
-    if (!_sir_resettextstyles())
-        _sir_selflog("error: failed to reset text styles!");
+    (void)_sir_setcolormode(SIRCM_16);
```
aremmell, line 144: Not sure I like this one. What's the point in changing this code just for the sake of it, and then (void)'ing the return value out?

Also, this will break everything. There is a reason I had an explicit `_sir_resettextstyles` here: the default color mode (in static memory) is `SIRCM_16`, but the color buffers have not been generated.

If you look at `_sir_setcolormode`, it will only reset the color buffers if the mode changes, which it will not here.

aremmell (changes requested):

src/sirinternal.c
aremmell: (on the first init, or if the colors mode is never changed)

johnsonjh: @aremmell 

Ah, I'll leave that check then, that's my bad.

The reason for the change was CBMC making an observation on the logic of this function, which essentially is: 

* Returns true on success and false on failure.
* Checks for invalid arguments, returns false if so.
* Do stuff in the function, print a message if there is a problem.
* Return true.


So the possible failure in the function doesn't result in a different return value.

johnsonjh: Also, the void isn't necessary.  Coverity (and Oracle) will complain if you return a value and check it sometimes - but not all the time.  We use this just once here, so it won't matter in this case.  But that's why I usually do that.

aremmell (approved):

PR #265, opened 2023-09-07
don't fstat log files on every write, only every n

aremmell:
Should improve write perf since fstat() is such a heavy call. There's no need to check it on every write, since relative to the roll size, one write is barely a dent.


johnsonjh (commented):
++

aremmell:
> ++

@johnsonjh it says no approving review still... 🤷

johnsonjh (approved):
+10% on Linux 👍🏻

PR #273, opened 2023-09-08
macroize _sir_valid* functions in order to capture source info

aremmell:
* As-is, some errors will always appear to be originating from sirhelpers.c, which is not useful. Changing these to macros will allow us to capture __func__, __file__, __LINE__, and store that in the error information.
* Also fixed a minor bug in plugin loader test


johnsonjh (approved):
++

PR #275, opened 2023-09-08
reset tls data on cleanup; only copy tid/name if set

aremmell:
This will fix the case of cleanup/init in a loop, in which less than 1s elapses (time stamp doesn't get generated)


johnsonjh (approved):
Tested and confirmed working

PR #276, opened 2023-09-08
_sir_validstrnofail does not need to call __sir_validstr

aremmell:
This should give us a tiny perf gain, and eliminates a useless code path.


johnsonjh (approved):
👍🏻

PR #284, opened 2023-09-08
Support building with dpsprintf instead of system/libc printf functions

johnsonjh:
* Support building with *dpsprintf* (instead of system/libc printf functions)


include/sir/helpers.h
```diff
@@ -435,8 +439,8 @@ bool _sir_getchar(char* input);
  */
 # define _sir_snprintf_trunc(dst, size, ...) \
     do { \
-      volatile size_t _n = size; \
-      if (!snprintf(dst, _n, __VA_ARGS__)) { (void)_n; }; \
+      volatile size_t _n = (size_t)size; \
+      if (!snprintf(dst, (int)(size_t)_n, __VA_ARGS__)) { (void)_n; }; \
```
aremmell, line 443: `_n` is already `size_t`... why the cast?

include/sir/helpers.h
```diff
@@ -29,6 +29,10 @@
 # include "sir/types.h"
 # include "sir/errors.h"
 
+# if defined(SIR_DPSPRINTF)
+#  include "sir/dpsprintf.h"
```
aremmell, line 33: Don't need this here; this header includes `sir/types.h` which includes `sir/platform.h` which already has this same block

aremmell (changes requested):

include/sir/helpers.h
johnsonjh: Aight.

include/sir/helpers.h
johnsonjh: I'll see.

johnsonjh: Ah, now I see, with the override function:

```sh
src/sirerrors.c:211:17: error: implicit conversion loses integer precision: 'volatile size_t' (aka 'volatile unsigned long') to 'int' [-Werror,-Wshorten-64-to-32]
                _sir_snprintf_trunc(heap_msg, SIR_MAXERROR, _SIR_E_PLATFORM_ERRORFORMAT, _sir_te.os_error,
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
include/sir/helpers.h:439:26: note: expanded from macro '_sir_snprintf_trunc'
      if (!snprintf(dst, _n, __VA_ARGS__)) { (void)_n; }; \
           ~~~~~~~~      ^~
src/sirerrors.c:216:9: error: implicit conversion loses integer precision: 'volatile size_t' (aka 'volatile unsigned long') to 'int' [-Werror,-Wshorten-64-to-32]
        _sir_snprintf_trunc(message, SIR_MAXERROR, SIR_ERRORFORMAT, _sir_te.loc.func, //-V576
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
include/sir/helpers.h:439:26: note: expanded from macro '_sir_snprintf_trunc'
      if (!snprintf(dst, _n, __VA_ARGS__)) { (void)_n; }; \
           ~~~~~~~~      ^~
2 errors generated.
```

It's just getting tossed away, so the precision doesn't matter - the easiest way to always silence the warning in every case is just to cast it to short.  I'll do that.

include/sir/helpers.h
```diff
@@ -436,7 +436,7 @@ bool _sir_getchar(char* input);
 # define _sir_snprintf_trunc(dst, size, ...) \
     do { \
       volatile size_t _n = size; \
-      if (!snprintf(dst, _n, __VA_ARGS__)) { (void)_n; }; \
+      if (!snprintf(dst, (short)_n, __VA_ARGS__)) { (void)_n; }; \
```
aremmell, line 439: we will get compiler warnings when using the standard snprintf if we do this i'm pretty sure. the portable one is `size_t` which isn't always the same type, in fact, it's almost never the same. It could be unsigned etc. You might want to do an #if defined() here, or at least test if casting to short in master causes a warning in CI

aremmell (changes requested):

include/sir/helpers.h
johnsonjh: It shouldn't ever cause a warning because the cast is explicit.  CI here has already these this - the DPSPRINTF isn't enabled yet on any of the build stage builds.  However, it will cause an warning if the mismatch is the other way, so technically, I should really cast it to a char =)  Did that in c5acfa0

johnsonjh: Actually this function is getting gross, I'll try to solve this a saner way.

johnsonjh:
I'll close this for now and reopen when I have time, but leave the branch.

PR #314, opened 2023-09-28
try to appease coverity by rearranging boundary checks

aremmell:
* Don't check idx, check sfc->count, which should teach it that the for statement can't run if idx is out of bounds


aremmell:
Don't merge yet. I think coverity is still going to barf on this.

aremmell:
Never mind. I think it's good.

johnsonjh (approved):

PR #340, opened 2023-10-17
Add in-place string mangling utilities

johnsonjh:
* Add in-place string mangling utilities

### DIFF
--- a/include/sir/utils.h
+++ b/include/sir/utils.h
@@ -1,0 +1,75 @@
+/**
+ * @file utils.h
+ *
+ * @brief String utility functions
+ *
+ * Miscellanous utility functions operating on NULL-terminated strings.
+ *
+ * @author    Jeffrey H. Johnson \<trnsz@pobox.com\>
+ * @date      2018-2023
+ * @version   2.2.4
+ * @copyright The MIT License (MIT)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+#ifndef _SIR_UTILS_H_INCLUDED
+# define _SIR_UTILS_H_INCLUDED
+
+# include "sir/platform.h"
+# include "sir/helpers.h"
+# include "sir/mutex.h"
+
+# if defined(__cplusplus)
+extern "C" {
+# endif
+
+/**
+ * @brief Remove all occurrences of substring "sub" in string "str".
+ */
+char* sir_strremove(char *str, const char *sub);
+
+/*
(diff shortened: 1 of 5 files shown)


src/sirutils.c
```diff
@@ -0,0 +1,117 @@
+/**
+ * @file sirutils.c
```
aremmell, line 2: 1. utils is a _very_ generic term to use for something specific to string manipulation. I would rename it `sirstrmanip` or something more explicit.
2. if you want these to be part of the public interface (i.e., don't begin with underscore), then they have to be fully Doxygen'd and the header included by `sir.h`. if you are unsure of your abilities, you can write synopsis for me in a text file or something and I can do the doxygen comments.

include/sir/utils.h
```diff
@@ -0,0 +1,74 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+#ifndef _SIR_UTILS_H_INCLUDED
+# define _SIR_UTILS_H_INCLUDED
+
+# include "sir/platform.h"
+# include "sir/helpers.h"
+# include "sir/mutex.h"
```
aremmell, line 35: does it really use something from helpers and mutex?

tests/tests.c
```diff
@@ -62,6 +62,7 @@ static sir_test sir_tests[] = {
     {"filesystem",              sirtest_filesystem, false, true},
     {"squelch-spam",            sirtest_squelchspam, false, true},
     {"plugin-loader",           sirtest_pluginloader, false, true},
+    {"strutils",                sirtest_strutils, false, true},
```
aremmell, line 65: recommend the pattern we've kind of established: `word-anotherword`. maybe `string-utils` or if you choose `strmanip` for the filename, `string-manip`

src/sirutils.c
```diff
@@ -0,0 +1,117 @@
...
+
+char* sir_strremove(char *str, const char *sub) {
+  if (!str)
+      return NULL;
+
+  const char* p;
+  char* r;
+  char* q;
+
+  if (!sub)
```
aremmell, line 40: seems like you could move this if-statement above those declarations; they will just be waste if sub is null

src/sirutils.c
```diff
@@ -0,0 +1,117 @@
...
+}
+
+char* sir_strredact(char *str, const char *sub, const char c) {
+  if (!str)
+      return NULL;
+
+  if (!sub)
+      return str;
+
+  char *p = strstr(str, sub);
```
aremmell, line 79: perhaps you want to include a case-insensitive version of anything that uses strstr? like using `strcasestr` (non-windows), and i can provide the windows equivalent or you can look in seatest

aremmell (dismissed):

src/sirutils.c
johnsonjh: Will do.

src/sirutils.c
johnsonjh: We'd need to have a strcasestr in impl since it's non-POSIX (AIX, Solaris, etc. don't have it.) but ... maybe.

src/sirutils.c
johnsonjh: I'll just move to helpers as discussed.

include/sir/utils.h
johnsonjh: It used to, but not now, good catch.

tests/tests.c
johnsonjh: I'll go with string-utils/stringutils

tests/tests.c
```diff
@@ -2009,6 +2010,85 @@ bool sirtest_pluginloader(void) {
...
+    _sir_eqland(pass, !_sir_strcreplace(NULL, 'c', 'n', -1));
+
+    TEST_MSG_0("_sir_strredact:   NULL pointer");
+    _sir_eqland(pass, !_sir_strredact(NULL, "s", '*'));
+
+    PASSFAIL_MSG(pass, "\t--- invalid string utility usage - NULL pointer: %s ---\n\n", PRN_PASS(pass));
+
+    TEST_MSG_0(WHITEB("--- invalid string utility usage - bad parameters ---"));
+
+    TEST_MSG_0("_sir_strremove:   bad parameter \"sub\"");
```
aremmell, line 2062: are you putting the extra spaces here and other spots on purpose? it's cool if so, just not sure if it's a typo.

johnsonjh: image

I could not include the spacing ...

src/sirhelpers.c
```diff
@@ -318,3 +318,90 @@ bool _sir_getchar(char* input) {
...
+  if (!sub)
+      return str;
+
+  char *p = strstr(str, sub);
+
+  if (!c || !p)
+      return str;
+
+  (void)memset(p, c, strnlen(sub, strlen(str)));
+
```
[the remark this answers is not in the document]
johnsonjh: Will change ...

aremmell (approved):

PR #346, opened 2023-10-23
Stricter SPDX compliance

johnsonjh:
* Stricter SPDX compliance
* Update README.md
* Remove loads of dead files from docs/*


bindings/ch/_sir_strremove.chf
```diff
@@ -1,13 +1,17 @@
...
  * Author:    Jeffrey H. Johnson <trnsz@pobox.com>
  * Copyright: Copyright (c) 2018-2023
  * Version:   2.2.4
  * License:   The MIT License (MIT)
  *
+ * -----------------------------------------------------------------------------
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Copyright (c) 2018-2023 Jeffrey H. Johnson <trnsz@pobox.com>
```
aremmell, line 13: It seems weird to duplicate this line essentially. There must be a way to combine them, even in the doxygen files

johnsonjh: Unfortunately I don't think there is.

If you mean the copyright line, it has to be that way to be recognized as valid license text, because it's part of the license, so it needs to be in the following format:
IMG_8414

SPDX and REUSE *do* allow you to completely get rid of the license text verbatim in each file, and use only a

If you want to do that, you then only need:

      SPDX-License-Identifier: MIT
      SPDX-FileCopyrightText: blah blah

If you don't use the full license text with a Copyright (c) .... then you should use the tag format, but if you have the full text it's not needed.  

As an aside, SPDX also allows you to distinguish between authors and other minor contributors too:
      SPDX-FileContributor: blah 
... which is where you'd list yourself if you took another file and modified it.

There are two "problems" though.  You cannot remove any license text or copyright notices that are already in the files from others (like the install script or those OpenBSD files we have).  So I decided to be consistent and have in every file three 'sections':

a) a prologue section which has the filename and possibly a description which may or may not be done with Doxygen tags.

b) any SPDX tags

c) the human readable license text.


---

REUSE/SPDX doesn't require a or c, but it does need it's specific format for b.

johnsonjh: Really, it would be nice if Doxygen supported SPDX and REUSE, but it doesn't seem to.

I looked at another project I use (RTEMS) that uses both REUSE and Doxygen, and they just don't use the license or copyright parts of Doxygen:

IMG_8415

bindings/python/sir.i
```diff
@@ -5,8 +5,13 @@
  * Copyright: Copyright (c) 2018-2023
  * Version:   2.2.4
  * License:   The MIT License (MIT)
+ *
+ * -----------------------------------------------------------------------------
+ *
  * SPDX-License-Identifier: MIT
  *
+ * Copyright (c) 2018-2023 Ryan M. Lederman <lederman@gmail.com>
```
aremmell, line 13: I don't want to have these duplicate copyright sections, whether this gives us "stricter" compliance or not... if there isn't a clean way to consolidate I would rather not change any file that already has copyright info either inline or in the .deps5

aremmell (changes requested):

bindings/python/sir.i
johnsonjh: Changed the syntax to avoid that …

example/example.c
```diff
@@ -5,10 +5,14 @@
  * libsir.
  *
  * @author    Ryan M. Lederman \<lederman@gmail.com\>
- * @date      2018-2023
+ * @date      Copyright (c) 2018-2023
```
aremmell, line 8: I really wish I could jive with this, but I just can't get down wit' it. If there is documentation in Doxygen that says you should put a Copyright in @date, then I'll reconsider. What did it say before, just @date 2018-2013???

aremmell (dismissed):

example/example.c
johnsonjh: SPDX and REUSE absolutely requires the Copyright (c) text somewhere.  If you don't want to put it there, then you'll have to duplicate it below that in the license area, like I had it before.

Or, you could do what all the other projects I've found that use Doxygen do that are SPDX/REUSE complaint (like FreeBSD, RTEMS, etc.) do, which is just not use those Doxygen fields at all, like ... 

image

The Doxygen documentation says that for the `@date`, `@copyright`, `@author` (and many others tags) that *"The text of the paragraph has no special internal structure"*, so there isn't any restriction on the text you can put there or how it's formatted.

johnsonjh: image

image

Etc

aremmell: I would rather lose the Doxygen fields then and do it the way FreeBSD does.

johnsonjh: Ok I'll do that later

johnsonjh: Done

utils/mcmb/mcmb.c
```diff
@@ -3011,3 +3018,5 @@ static CMB_TRANSFORM_OP_FIND(*, cmb_mul_find)
 static CMB_TRANSFORM_OP_FIND(/, cmb_div_find)
 static CMB_TRANSFORM_OP_FIND(+, cmb_add_find)
 static CMB_TRANSFORM_OP_FIND(-, cmb_sub_find)
+
+// vim: filetype=c:tabstop=4:ai:expandtab
```
aremmell, line 3022: Accident?

aremmell (approved):
All good if you review my comments and they jive

utils/mcmb/mcmb.c
johnsonjh: Nope, that's a vi modeline.  They have to be somewhere within the first 5 or last 5 lines.  I didn't want it at the top messing up the same style we use everywhere else.

bindings/ch/sir_output.chi
```diff
@@ -67,3 +70,5 @@ SIR_OUTPUT(const char* message...) {
 
     return retval;
 }
+
+// vim: filetype=ch:autoindent:expandtab
```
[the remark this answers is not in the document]
johnsonjh: It is because just this one file is ending in .chi ... it makes it get treated the same as .ch and .chf

bindings/ch/sir_output.chi
```diff
@@ -1,13 +1,13 @@
 /*
- * SPDX-License-Identifier: MIT
- *
  * SIR_OUTPUT.chf
```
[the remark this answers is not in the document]
johnsonjh: No, it's OK: that's a template.  The `SIR_OUTPUT` part gets replaced by the generated function.  When you make the bindings you'll see the:
```
[binding] generated sir_alert.chf.
[binding] generated sir_crit.chf.
[binding] generated sir_debug.chf.
[binding] generated sir_emerg.chf.
[binding] generated sir_error.chf.
[binding] generated sir_info.chf.
[binding] generated sir_notice.chf.
[binding] generated sir_warn.chf.
```
Each generated file gets the right names :)

Doxyfile
```diff
@@ -1,7 +1,35 @@
 # Doxyfile 1.9.8
+
+################################################################################
```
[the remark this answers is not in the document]
johnsonjh: It's 80 Columbus, and they're the same 
IMG_8744

.vscode/libsir.code-snippets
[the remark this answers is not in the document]
johnsonjh: Yeah, they're the right lengths:)

example/example.c
```diff
@@ -25,11 +29,18 @@
...
+ * -----------------------------------------------------------------------------
  */
+
 #include "sir.h"
 #include "sir/helpers.h"
 #include "sir/internal.h"
 
+#if defined(__EMSCRIPTEN__)
+# include <emscripten/emscripten.h>
+#endif
```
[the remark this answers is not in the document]
johnsonjh: Oh, example doesn't need it now - I switch to linking with a different option, I can remove that now.

tests/tests.h
```diff
@@ -34,6 +40,10 @@
 # include "sir/threadpool.h"
 # include "sir/queue.h"
 
+# if defined(__EMSCRIPTEN__)
```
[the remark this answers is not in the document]
johnsonjh: I actually don't need it anymore - I did temporarily when I had to call emplatform_exit or something stupid.

aremmell (approved):

PR #361, opened 2023-11-27
Add support for CompCert-C and disabling shared library support.

johnsonjh:
* Add just enough support for building with CompCert-C and Chamois-CompCert (with the caveat that multithreading is working but lacks TLS and is not officially supported on this toolchain).  Tested on Intel x86_64 and Kalray KVX.
* Allow the user to define _sir_thread_local externally.
* Update README.md to note that Kalray MPPA/Coolidge (KVX) tested (with QEMU KVX) and working.
* Add Chamois-CompCert to CI


Makefile
```diff
@@ -298,9 +312,9 @@ $(OUT_SHARED): $(OBJ_SHARED)
 
 static: $(OUT_STATIC)
 
-$(OUT_STATIC): $(OUT_SHARED)
```
aremmell, line 301: Won't this break the dependency tree? The static library depends on the shared library objects to create the archive. Did you test it both ways?

README.md
```diff
@@ -104,7 +104,7 @@ At this time, libsir is supported (*that is, it compiles and passes the test sui
 | **SerenityOS** | **GCC**&nbsp;(13.1) |
 | **WebAssembly/JS** | **Emscripten**&nbsp;(3.1.45&nbsp;‑&nbsp;3.1.48),&nbsp; **Node.js**&nbsp;(20.5&nbsp;‑&nbsp;21.2) |
 
-* libsir is known to work on most common (and many uncommon) architectures. It has been built and tested on **Intel** (x86\_64, i686, x32), **ARM** (ARMv6, ARMv7HF, ARMv8‑A), **POWER** (PowerPC, PPC64, PPC64le), **MIPS** (MIPS64, MIPS32, 74Kc), **SPARC** (SPARC64, V8, LEON3), **z/Architecture** (S390X), **SuperH** (SH‑4A), **RISC‑V** (RV64), **OpenRISC** (OR1200), and **m68k** (68020+).
+* libsir is known to work on most common (and many uncommon) architectures. It has been built and tested on **Intel** (x86\_64, i686, x32), **Kalray MPPA®** (Coolidge™, KVX), **ARM** (ARMv6, ARMv7HF, ARMv8‑A), **POWER** (PowerPC, PPC64, PPC64le), **MIPS** (MIPS64, MIPS32, 74Kc), **SPARC** (SPARC64, V8, LEON3), **z/Architecture** (S390X), **SuperH** (SH‑4A), **RISC‑V** (RV64), **OpenRISC** (OR1200), and **m68k** (68020+).
```
aremmell, line 107: Let's move this toward the end of the list since it's fringe/esoteric

include/sir/platform.h
```diff
@@ -613,20 +613,22 @@ typedef BOOL(CALLBACK* sir_once_fn)(PINIT_ONCE, PVOID, PVOID*);
...
-#  endif
-# elif defined(__WIN__)
-#  define _sir_thread_local __declspec(thread)
-# elif defined(__GNUC__) || (defined(_AIX) && (defined(__xlC_ver__) || defined(__ibmxl__)))
-#  define _sir_thread_local __thread
-# else
-#  if !defined(_CH_) && !defined(__CH__)
-#   error "unable to resolve thread local attribute; please contact the author."
+#   if !defined(_CH_) && !defined(__CH__)
+#    error "unable to resolve thread local attribute; please contact the author."
```
aremmell, line 630: We should change this and the other messages like it to say 'contact the maintainers'

Makefile
```diff
@@ -78,7 +79,7 @@ SIR_XSTD ?= -std=c++20
 # Base CFLAGS
 
 ifneq ($(NO_DEFAULT_CFLAGS),1)
-  SIR_CFLAGS += -Wall -Wextra $(strip $(strip $(WPEDANTIC) -Iinclude) $(SIR_FPIC))
+  SIR_CFLAGS += -Wall $(strip $(WARNEXTRA) $(WPEDANTIC) $(SIR_FPIC) -Iinclude)
```
aremmell, line 82: How come?

Makefile
```diff
@@ -116,6 +117,13 @@ ifeq ($(SIR_NO_SYSTEM_LOGGERS),1)
   SIR_CFLAGS += -DSIR_NO_SYSTEM_LOGGERS
 endif
 
+##############################################################################
+# Disable shared library support?
+
+ifeq ($(SIR_NO_SHARED),1)
```
aremmell, line 123: I thought the goal was to create a build that doesn't depend on pthreads. How did we get to no shared libraries, and why?

johnsonjh: Kalray's toolchain (and CompCert-C) doesn't support building shared libraries from the compilers frontend.

Also, the make install would fail due to missing shared libraries if you only built static.  I expect this to mostly be set by the build (like the CompCert section of sirplatform.mk) and not by users directly.

I still want to make a build that doesn't depend on anything from pthreads, but this is mostly unrelated to that since I've been able to (mostly) work around my need for that, for now.

sirplatform.mk
```diff
@@ -282,6 +282,23 @@ ifeq ($(AIXTLS),1)
   SIR_XFLAGS+=-fPIC -ftls-model=global-dynamic
 endif
 
+##############################################################################
+# CompCert (WIP)
```
aremmell, line 286: If it's WIP, why PR?  Also, forgot to mention it before, but if we're adding a preproc macro, we need to update README (iirc) and `docs/sources/config.md`

aremmell (commented):
Just need to add relevant changes to docs (at least docs/sources/config.md, but I think README.md too)

sirplatform.mk
johnsonjh: I put WIP there because it needs the unreleased version from the vendors GitLab repo to build out of the box.  I'll add a better comment and update the documentation.

johnsonjh: Done.

aremmell (approved):

PR #365, opened 2023-12-11
Add experimental support for PlatformIO, just to see if we can get it to work on an ESP32 or similar.

aremmell:
- Add new macro `SIR_NO_TEXT_STYLING` which disables all text styling and color-related code.
- Add new macro `SIR_EMBEDDED`, which if defined before `make`, will set the following:
  - `SIR_NO_SYSTEM_LOGGERS`=`1`
  - `SIR_NO_PLUGINS`=`1`
  - `SIR_SELFLOG`=`0`
  - `SIR_NO_TEXT_STYLING`=`1`
- `SIR_EMBEDDED` will also be defined by `platform.h` if `PLATFORMIO` or `ARDUINO` are defined. The above macros will then be defined as described.
- Stub out some functions/typedefs that are not available on the target embedded platform(s).
- Shrink fixed buffer sizes (max message length, hostname, buffers that will be rendered useless by preprocessor) to be more appropriate for embedded platforms.


johnsonjh (dismissed):

johnsonjh (approved):
👍🏻

PR #367, opened 2023-12-13
Use hexadecimal for printing numeric thread IDs on NetBSD; misc. CI updates

johnsonjh:
* Use hexadecimal for printing numeric thread IDs on NetBSD
* Fixes for running 32-bit binaries in CI container environment


include/sir/config.h
```diff
@@ -93,12 +93,23 @@
 # define SIR_PIDSUFFIX ")"
 
 /**
- * The format for the current process/thread ID.
+ * The format for the current process ID.
  *
  * @remark Only applies if ::SIRO_NOPID or ::SIRO_NOTID are not set.
```
aremmell, line 98: If you split them then this comment is inaccurate now; should just be NOPID

include/sir/config.h
```diff
@@ -93,12 +93,23 @@
...
  */
 # define SIR_PIDFORMAT "%d"
 
+/**
+ * The format for the current thread ID.
+ *
+ * @remark Only applies if ::SIRO_NOPID or ::SIRO_NOTID are not set.
```
aremmell, line 105: Ditto for this one

aremmell (approved):
We are gonna have merge conflicts but 🤷🏻‍♂️

include/sir/config.h
johnsonjh: I'll fix

include/sir/config.h
```diff
@@ -93,12 +93,23 @@
...
+ * @remark Only applies if ::SIRO_NOPID is not set.
  */
 # define SIR_PIDFORMAT "%d"
 
+/**
+ * The format for the current thread ID.
+ *
+ * @remark Only applies if ::SIRO_NOTID is not set.
+ */
+# if defined(__NetBSD__)
```
aremmell, line 107: why don't we change this to `__USE_HEX_TIDS__`, and define it in platform so we don't have to put all the platforms in this header? thoughts?

thinking ahead to when we have 4-5 here

johnsonjh: Sure ...

aremmell (approved):

PR #376, opened 2023-12-26
Tweaks based on cppcheck linter warnings

aremmell:
- variable `foo` can be declared as pointer to const
- local variable `foo` shadows outer function
- the scope of the variable `foo` can be reduced
- condition `foo` is always true/false
- function argument 2 names different
- variable `foo` is assigned a value that is never used
- variable `foo` is reassigned a value before the old one has been used

### DIFF
--- a/tests/tests.c
+++ b/tests/tests.c
@@ -1773,11 +1773,9 @@ bool sirtest_filesystem(void) {
         1234
     };
 
-#if defined(__WIN__)
     if (get_wineversion()) {
         bad_fds[3] = 0;
     }
-#endif
 
     for (size_t n = 0; n < _sir_countof(bad_fds); n++) {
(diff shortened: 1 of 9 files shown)


tests/tests.c
```diff
@@ -1775,9 +1773,11 @@ bool sirtest_filesystem(void) {
         1234
     };
 
+#if defined(__WIN__)
     if (get_wineversion()) {
```
johnsonjh, line 1777: I don't think this is correct - we won't be running Wine on Windows.

johnsonjh (approved):

aremmell (commented):
👍

johnsonjh (approved):

johnsonjh (dismissed):
👍🏻

johnsonjh (approved):
😑

PR #390, opened 2024-02-05
Fix `mcmb` on IBM i (OS/400) PASE; fix DMD ImportC 2.107.0 incompatibility

johnsonjh:
* Fix IBM i (OS/400) PASE `mcmb` compatibility
* Update to DMD ImportC 2.107.0 and workaround a bug


src/sirinternal.c
```diff
@@ -55,10 +55,17 @@ static sirconfig _sir_cfg      = {0};
 static sirfcache _sir_fc       = {0};
 static sir_plugincache _sir_pc = {0};
 
+#if !defined(__IMPORTC__)
 static sir_mutex cfg_mutex  = SIR_MUTEX_INIT;
 static sir_mutex fc_mutex   = SIR_MUTEX_INIT;
 static sir_mutex pc_mutex   = SIR_MUTEX_INIT;
 static sir_mutex ts_mutex   = SIR_MUTEX_INIT;
+#else
```
aremmell, line 63: just do this where SIR_MUTEX_INIT is defined, silly.

tests/tests.c
```diff
@@ -1099,7 +1099,11 @@ bool sirtest_mutexsanity(void) {
     TEST_MSG_0(WHITEB("create, lock, unlock, destroy"));
     (void)printf(INDENT_ITEM WHITE("creating mutex...") "\n");
 
+#if !defined(__IMPORTC__)
     sir_mutex m1 = SIR_MUTEX_INIT;
+#else
```
aremmell, line 1104: Same

aremmell (changes requested):

src/sirinternal.c
johnsonjh: That's the bug tho, as discussed.

tests/tests.c
johnsonjh: =)

aremmell (approved):

PR #404, opened 2024-03-09
Add `SIR_NO_TEXT_STYLING` build option to disable ANSI output styling; add `SIR_USE_EOL_CRLF` build option to use `CR`+`LF` line endings, instead of `LF`; switch Anaconda CI container from Anaconda to conda-forge Python; update Intel oneAPI DPC++/C++ to 2024.0.3 (via conda); update tested compiler versions and some minor cosmetic cleanups; drop any unnecessary use of `ccache` in CI containers; update Android NDK to latest prerelease build; bump upper Android API level from 33 to 35

### LINKED ISSUE #405: Update Intel C to 2024.0.3+
Changelog is out ... but repos are still serving 2024.0.0 or 2024.0.1 currently.

johnsonjh:
* Add `SIR_NO_TEXT_STYLING` build option to disable ANSI output styling
* Add `SIR_USE_EOL_CRLF` build option to use `CR`+`LF` line endings, instead of `LF`
* Switch Anaconda CI container from Anaconda to conda-forge Python
* Update Intel oneAPI DPC++/C++ to 2024.0.3 (via conda)
  * Closes #405 
* Update tested compiler versions and some minor cosmetic cleanups
* Drop any unnecessary use of `ccache` in CI containers
* Update Android NDK to latest prerelease build
* Bump upper Android API level from 33 to 35


include/sir/ansimacros.h
```diff
@@ -68,68 +82,278 @@
...
  * Ends by resetting to the default fg/bg color, and normal attr.
  */
-# define COLOR(attr, fg, bg, s) \
-    SIR_ESC_SEQ(#attr ";" #fg ";" #bg, s) SIR_ESC_SEQE("0;39;49")
+# undef COLOR
+# if !defined(SIR_NO_TEXT_STYLING)
+#  define COLOR(attr, fg, bg, s) \
+     SIR_ESC_SEQ(#attr ";" #fg ";" #bg, s) SIR_ESC_SEQE("0;39;49")
+# else
+#  define COLOR(attr, fg, bg, s) s
```
aremmell, line 90: Are we going to get unreferenced variable for the other args? I forget if macros do that or not. (do we need SIR_UNUSED(attr), etc)?

include/sir/ansimacros.h
```diff
@@ -68,68 +82,278 @@
...
+#  define BREDB(s) COLOR(1, 91, 49, s) /**< Bold bright red foreground text. */
+# else
+#  define BREDB(s) s
+# endif
+
+# undef GREEN
+# if !defined(SIR_NO_TEXT_STYLING)
+#  define GREEN(s) COLOR(0, 32, 49, s) /**< Green foreground text. */
+# else
+#  define GREEN(s) s
```
aremmell, line 181: You're going to hate me but I would like this better as a huge:

```c
#if !defined(SIR_NO_TEXT_STYLING)
 #define MACRO1(..., s) ...
 #define MACRO2(..., s) ...
 ...
#else
 #define MACRO1(..., s) s
 #define MACRO2(..., s) s
 ...
#endif
```
Rather than a bunch of #ifdefs

include/sir/config.h
```diff
@@ -373,8 +373,14 @@
  * data in any color mode (the largest possible sequence, which is:
  * `\x1b[a;fb;m;rrr;ggg;bbb;fb;m;rrr;ggg;bbbm`) plus a null terminator.
  */
-# if !defined(SIR_MAXSTYLE)
-#  define SIR_MAXSTYLE 43
+# if !defined(SIR_NO_TEXT_STYLING)
```
aremmell, line 376: if `SIR_EOL_CRLF` is defined we need to add 1 to `SIR_MAXSTYLE`

src/sirinternal.c
```diff
@@ -59,12 +59,16 @@ static sir_plugincache _sir_pc = {0};
 static sir_mutex cfg_mutex  = SIR_MUTEX_INIT;
 static sir_mutex fc_mutex   = SIR_MUTEX_INIT;
 static sir_mutex pc_mutex   = SIR_MUTEX_INIT;
+# if !defined(SIR_NO_TEXT_STYLING)
 static sir_mutex ts_mutex   = SIR_MUTEX_INIT;
+# endif
 #else
 static sir_mutex cfg_mutex  = {0};
```
aremmell, line 66: tbh, i would rather set all of these to `= {0}` here, OR if we can get away with no warnings, just don't initialize them at all and delete `SIR_MUTEX_INIT`...

Since on everything but Windows `SIR_MUTEX_INIT` is `PTHREAD_MUTEX_INITIALIZER` which expands to `#define PTHREAD_MUTEX_INITIALIZER {_PTHREAD_MUTEX_SIG_init, {0}}` and `_PTHREAD_MUTEX_SIG_init` is a magic number, and it is still working with importc and pthreads, it may not even be necessary to use `PTHREAD_MUTEX_INITIALIZER` at all, so these can always just be `= {0}` and `SIR_MUTEX_INIT` can go adios

src/sirinternal.c
```diff
@@ -197,7 +203,11 @@ bool _sir_init(sirinit* si) {
 
     _SIR_UNLOCK_SECTION(SIRMI_CONFIG);
 
+#if defined(SIR_NO_TEXT_STYLING) && defined(SIR_NO_SYSTEM_LOGGERS)
+    _sir_selflog("initialized successfully");
```
aremmell, line 207: not needed; `init` is initialized to true, so it will say successfully already

src/sirinternal.c
```diff
@@ -471,8 +483,13 @@ bool _sir_mapmutexid(sir_mutex_id mid, sir_mutex** m, void** section) {
             tmpsec = &_sir_pc;
             break;
         case SIRMI_TEXTSTYLE:
+#if !defined(SIR_NO_TEXT_STYLING)
             tmpm   = &ts_mutex;
             tmpsec = &sir_text_style_section;
+#else
```
aremmell, line 489: I don't know what the side effects of this will be unless we have also removed all `_sir_locksection` (`_SIR_LOCK_SECTION`) for `SIRMI_TEXSTYLE`, so we better check that

src/sirinternal.c
aremmell, line 1: Comment at 671/692 should be removed

src/sirinternal.c
aremmell: Worth a shot. We will know rather quickly if `pthread_create` fails.

src/sirtextstyle.c
```diff
@@ -215,3 +217,43 @@ bool _sir_setcolormode(sir_colormode mode) {
 
     return true;
 }
+#else /* SIR_NO_TEXT_STYLING */
+const char* _sir_gettextstyle(sir_level level) {
+    SIR_UNUSED(level);
```
aremmell, line 222: I think it would be better to also `SIR_ASSERT(false)` before returning NULL, in case some naive caller doesn't expect a NULL retval. Or alternatively, return `""`.

src/sirtextstyle.c
```diff
@@ -215,3 +217,43 @@ bool _sir_setcolormode(sir_colormode mode) {
...
+}
+
+bool _sir_settextstyle(sir_level level, const sir_textstyle* style) {
+    SIR_UNUSED(level);
+    SIR_UNUSED(style);
+    return false;
+}
+
+const sir_textstyle* _sir_getdefstyle(sir_level level) {
+    SIR_UNUSED(level);
```
aremmell, line 233: `SIR_ASSERT(false)` or return `""` here too

src/sirinternal.c
johnsonjh: Yes, but PVS-Studio will complain if both SIR_NO_TEXT_STYLING and SIR_NO_SYSTEM_LOGGERS are defined that the condition is always true, because there is no way that init could be set to false.  

But I can do it that way and suppress the warning, if you prefer.

src/sirinternal.c
johnsonjh: OK.

src/sirtextstyle.c
johnsonjh: Let's SIR_ASSERT(false);

src/sirtextstyle.c
johnsonjh: SIR_ASSERT(false)'ing.

src/sirinternal.c
johnsonjh: It gave me warnings ... I'll mess with this in another PR.

include/sir/config.h
johnsonjh: Yep.

src/sirinternal.c
johnsonjh: I checked it ... and verified that's the case now.

example/example.c
```diff
@@ -194,7 +194,7 @@ int main(void) {
 int report_error(void) {
     char message[SIR_MAXERROR] = {0};
     (void)sir_geterror(message);
-    (void)fprintf(stderr, "\x1b[31mlibsir error: %s\x1b[0m\n", message);
+    (void)fprintf(stderr, BREDB("libsir error: ") RED("%s") "\r\n", message);
```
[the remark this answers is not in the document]
johnsonjh: Out of habit, because DPS8M operates in raw mode, where it's needed.  It's not needed here tho.

include/sir/ansimacros.h
johnsonjh: No, didn't get any warnings, it's OK with that for macros.

include/sir/ansimacros.h
johnsonjh: *grumble grumble*

let me look at it.

johnsonjh: OK done.

plugins/sample/plugin_sample.c
```diff
@@ -63,22 +63,22 @@ PLUGIN_EXPORT bool sir_plugin_query(sir_plugininfo* info) {
     info->desc      = desc;
     info->caps      = caps;
 
-    (void)printf("\t" DGRAY("plugin_sample ('%s')") "\n", __func__);
+    (void)printf("\t" SIR_DGRAY("plugin_sample ('%s')") "\n", __func__);
```
aremmell, line 66: technically, if we're going to have that CRLF setting, we should have a `SIR_LE` macro for line endings so we can clean these up too

src/sirhelpers.c
```diff
@@ -105,10 +105,10 @@ bool __sir_validupdatedata(const sir_update_config_data* data, const char* func,
     if (valid && _sir_bittest(data->fields, SIRU_SYSLOG_CAT))
         valid = _sir_validstrnofail(data->sl_category);
 
-    if (!valid) {
+    if (!valid) { // GCOVR_EXCL_START
```
aremmell, line 108: what's this aboot? it would be very easy to make a test to cover that unless the setter function won't let you set an empty string

src/sirinternal.c
aremmell: ya rite, we will forget.

src/sirinternal.c
```diff
@@ -613,12 +628,14 @@ bool _sir_logv(sir_level level, PRINTF_FORMAT const char* format, va_list args)
     buf.pid       = cfg.state.pidbuf;
     buf.name      = cfg.si.name;
 
+#if !defined(SIR_NO_TEXT_STYLING)
```
aremmell, line 631: we should verify that this is not making an assumption that `buf.style` will be filled with a valid string (or i suppose maybe it doesn't matter if that part is preprocessed out as well)

src/sirplugins.c
```diff
@@ -255,11 +255,11 @@ bool _sir_plugin_unload(sir_plugin* plugin) {
                      " cleanup!", plugin->path, plugin->handle);
 
 # if !defined(__WIN__)
-    if (0 != dlclose(plugin->handle)) {
+    if (0 != dlclose(plugin->handle)) { // GCOVR_EXCL_START
```
aremmell, line 258: when you add these, do you brace yourself for the inevitable comment?

src/sirplugins.c
```diff
@@ -343,10 +343,10 @@ sirpluginid _sir_plugin_cache_add(sir_plugincache* spc, sir_plugin* plugin) {
     if (!_sir_validptr(spc) || !_sir_validptr(plugin))
         return 0U;
 
-    if (spc->count >= SIR_MAXPLUGINS) {
+    if (spc->count >= SIR_MAXPLUGINS) { // GCOVR_EXCL_START
```
aremmell, line 346: samesies

src/sirtextstyle.c
```diff
@@ -203,15 +205,57 @@ bool _sir_setcolormode(sir_colormode mode) {
         /* when the color mode changes, it's necessary to regenerate the text styles
          * we're holding. for example in the case of downgrading color modes, the
          * styles in memory could be incompatible with the new mode. */
-        if (!_sir_resettextstyles()) {
+        if (!_sir_resettextstyles()) { // GCOVR_EXCL_START
```
aremmell, line 208: 💀

utils/mcmb/mcmb.c
```diff
@@ -1382,7 +1382,6 @@ static const char
   return ret;
 }
 
-
```
aremmell, line 1385: oooh how'd this get by?

aremmell (changes requested):

utils/mcmb/mcmb.c
johnsonjh: ☠️

src/sirplugins.c
johnsonjh: 💀 but no way to force the test really for this path :(

src/sirplugins.c
johnsonjh: 💀 I guess you could try to load 625282 plug-ins?

src/sirhelpers.c
johnsonjh: I tried and I couldn't get any coverage here consistently...  I'm not sure why

src/sirinternal.c
johnsonjh: I will check

plugins/sample/plugin_sample.c
johnsonjh: I can do that

src/sirtextstyle.c
johnsonjh: 💀

plugins/sample/plugin_sample.c
johnsonjh: Done.

src/sirinternal.c
johnsonjh: This is the only place I see buf.style actually

src/sirinternal.c
```diff
@@ -833,10 +850,10 @@ const char* _sir_format(bool styling, sir_options opts, sirbuf* buf) {
         if (styling)
             (void)_sir_strncat(buf->output, SIR_MAXOUTPUT, SIR_ESC_RST, SIR_MAXSTYLE);
 
-#if defined(SIR_EOL_CRLF)
-        (void)_sir_strncat(buf->output, SIR_MAXOUTPUT, "\r\n", 2);
```
aremmell, line 837: This can be consolidated to use `SIR_EOL` now

src/sirinternal.c
```diff
@@ -1534,11 +1551,11 @@ long __sir_nprocs(bool test_mode) {
         nprocs = vtprocs;
 #endif
 
-    if (nprocs < 1) {
-        _sir_selflog(BRED("Failed to determine processor count!"));
+    if (nprocs < 1) { // GCOVR_EXCL_START
```
aremmell, line 1554: adding GCOVR_EXCL_START doesn't make us have more code coverage, it just makes it look like we do. it's stupid.

src/sirplugins.c
aremmell: Well, it would probably be less ghetto to add some macro like `SIR_PURPOSEFUL_ERRORS` and put code inside that. For example, here, set the handle to null or unload the module before the test. It's cleaner (and also tests the code path) than putting these comments everywhere that GCOVR just happens to never hit.

src/sirplugins.c
aremmell: I have no problem loading 17 plugins over pretending we have coverage here.

src/sirtextstyle.c
aremmell: Easy to break this too. Simply make `_sir_formatstyle()` fail.

src/sirtextstyle.c
```diff
@@ -203,15 +205,57 @@ bool _sir_setcolormode(sir_colormode mode) {
...
     return true;
 }
+#else /* SIR_NO_TEXT_STYLING */
+bool _sir_settextstyle(sir_level level, const sir_textstyle* style) {
+    SIR_UNUSED(level);
+    SIR_UNUSED(style);
+    return false;
+}
+
+const char* _sir_gettextstyle(sir_level level) { // GCOVR_EXCL_START
```
aremmell, line 227: This is a legitimate use of GCOVR_EXCL_START. This and the default case in a switch are pretty much the only kind that are though.

tests/tests++.cc
```diff
@@ -96,7 +96,7 @@ int main(int argc, char** argv) {
 
         print_test_summary(tgt_tests, passed, sir_timer_elapsed(&timer));
 
-        if (passed != tgt_tests) {
+        if (passed != tgt_tests) { // GCOVR_EXCL_START
```
aremmell, line 99: Artificially set passed to passed -1 when `--coverage` is on the CL?

tests/tests.c
```diff
@@ -218,9 +218,9 @@ bool sirtest_logwritesanity(void) {
         _sir_eqland(pass, found);
 
         if (found)
-            TEST_MSG(GREEN("found '%s'"), message);
+            TEST_MSG(SIR_GREEN("found '%s'"), message);
         else
-            TEST_MSG(RED("did not find '%s'"), message);
+            TEST_MSG(SIR_RED("did not find '%s'"), message); // GCOVR_EXCL_LINE
```
aremmell, line 223: `buf[0] = '\0'`

tests/tests.c
```diff
@@ -1065,24 +1069,24 @@ bool sirtest_levelssanity(void) {
...
+                                 " %04"PRIx16) ")" SIR_EOL, n + 1, iterations, rand_count, levels);
     }
-    PASSFAIL_MSG(pass, "\t--- random bitmask of valid levels: %s ---\n\n", PRN_PASS(pass));
+    PASSFAIL_MSG(pass, "\t--- random bitmask of valid levels: %s ---" SIR_EOL SIR_EOL, PRN_PASS(pass));
 
-    TEST_MSG_0(WHITEB("--- invalid values ---"));
+    TEST_MSG_0(SIR_WHITEB("--- invalid values ---"));
 
     /* greater than SIRL_ALL. */
     sir_levels invalid = (0xffffu & ~SIRL_ALL);
```
aremmell, line 1080: We use an upper case `U` suffix down below a few lines... one of these is wrong. I don't remember which one we use more of but they should all be the same...

tests/tests.c
```diff
@@ -2019,9 +2023,9 @@ bool sirtest_stringutils(void) {
     INIT(si, SIRL_ALL, 0, 0, 0);
     bool pass = si_init;
 
-    char str[] = "Kneel  \f \n  before  \t \r \v  Zod!?";
+    char str[] = "Kneel  \f \x0a  before  \t \x0d \v  Zod!?";
```
aremmell, line 2026: Y

tests/tests_shared.c
```diff
@@ -33,39 +33,39 @@
...
-    (void)printf(WHITEB("\n" ULINE("libsir") " %s (%s) running %zu %s...") "\n",
+    (void)printf(SIR_WHITEB(SIR_EOL SIR_ULINE("libsir") " %s (%s) running %zu %s...") SIR_EOL,
         sir_getversionstring(), (sir_isprerelease() ? "prerelease" : "release"),
         tgt_tests, TEST_S(tgt_tests));
 }
 
 void print_test_intro(size_t num, size_t tgt_tests, const char* const name) {
-    (void)printf(WHITEB("\n(%zu/%zu) '%s'...") "\n\n", num, tgt_tests, name);
+    (void)printf(SIR_EOL SIR_WHITEB("(%zu/%zu) '%s'...") SIR_EOL SIR_EOL, num, tgt_tests, name);
 }
```
aremmell, line 44: These 3 are changing the order of the newline/color start token. Should either not do that or run it and make sure it does not change the look of the output

tests/tests_shared.c
```diff
@@ -215,12 +215,12 @@ bool parse_cmd_line(int argc, char** argv, const sir_cl_arg* args, size_t num_ar
     return true;
 }
 
-void wait_for_keypress(void) {
-    (void)printf(WHITEB(EMPH("press any key to exit...")) "\n");
+void wait_for_keypress(void) { // GCOVR_EXCL_START
```
aremmell, line 218: Why do I recall we had already resolved this? You wrote a script that called it with --wait and then killed it or something like that

tests/tests_shared.c
```diff
@@ -215,12 +215,12 @@ bool parse_cmd_line(int argc, char** argv, const sir_cl_arg* args, size_t num_ar
...
+    (void)printf(SIR_WHITEB(SIR_EMPH("press any key to exit...%s")), SIR_EOL);
```
aremmell, line 219: We need to be consistent with the `SIR_EOL` in the `printf`s. In some you add them after the format string, and in some you add %s and make them an argument. They should all be identical.

tests/tests_shared.c
```diff
@@ -261,7 +261,7 @@ size_t sir_readline(FILE* f, char* buf, size_t size) {
 
     while (idx < size) {
         int ch = getc(f);
-        if (EOF == ch || '\n' == ch)
+        if (EOF == ch || '\x0a' == ch || '\x0d' == ch)
```
aremmell, line 264: Y

tests/tests_shared.h
```diff
@@ -113,27 +113,32 @@ extern "C" {
...
 /** Prints `msg` in red to stderr. */
-# define ERROR_MSG(msg, ...) TEST_MSG(RED(msg), __VA_ARGS__)
+# define ERROR_MSG(msg, ...) TEST_MSG(SIR_RED(msg), __VA_ARGS__)
 
 /** ERROR_MSG but no varargs. Use when `msg` is just a string. */
-# define ERROR_MSG_0(msg) TEST_MSG_0(RED(msg))
+# define ERROR_MSG_0(msg) TEST_MSG_0(SIR_RED(msg))
 
 /** Prints `msg` in green if `expr` is true, or in red if false. */
-# define PASSFAIL_MSG(expr, msg, ...) \
```
aremmell, line 135: Aren't `SIR_GREEN` and `SIR_RED` both just the input arg when `SIR_NO_TEXT_STYLING`?

aremmell (changes requested):

tests/tests_shared.c
johnsonjh: Just did a universal search and replace across all files for \r and \n and replaced with proper macro and/or \x0a or \x0d, repeating until there was no more \r or \n.  On this one particular, I check for both because either can be used to end the line depending on the terminal type (mostly old stuff like A/UX using on \r though).  I'll keep \r \n if you prefer tho.

tests/tests_shared.h
johnsonjh: If you allow it to be always true or false, you get one of those always true and/or always false warnings that you'd have to suppress otherwise.

tests/tests_shared.c
johnsonjh: I did these all automatically with a macro, where it did the easiest thing for the most part.  In this case, it doesn't make sense to have `'COLOR("message\n")'` when we should be using `'COLOR("message")\n'` -- we shouldn't be coloring the newline, so I'll fix that.

tests/tests_shared.c
johnsonjh: Fixed

tests/tests_shared.c
johnsonjh: This is a case where I think the macro did the right thing.  We shouldn't be coloring invisible newlines.  The color should start and end with the text to be colorized.  I did run it and it does look the same.

src/sirerrors.c
```diff
@@ -176,7 +176,7 @@ bool __sir_handlewin32err(DWORD code, const char* func, const char* file, uint32
 
     DWORD fmtmsg = FormatMessageA(flags, NULL, code, 0, (LPSTR)&errbuf, SIR_MAXERROR, NULL);
     if (0 < fmtmsg && _sir_validstrnofail(errbuf)) {
-        if (errbuf[fmtmsg - 1] == '\n' || errbuf[fmtmsg - 1] == ' ')
+        if (errbuf[fmtmsg - 1] == '\x0a' || errbuf[fmtmsg - 1] == ' ')
```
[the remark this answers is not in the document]
johnsonjh: Changed.

plugins/sample/plugin_sample.c
```diff
@@ -63,22 +63,22 @@ PLUGIN_EXPORT bool sir_plugin_query(sir_plugininfo* info) {
...
 }
 
 PLUGIN_EXPORT bool sir_plugin_init(void) {
-    (void)printf("\t" DGRAY("plugin_sample ('%s')") "\n", __func__);
+    (void)printf("\t" SIR_DGRAY("plugin_sample ('%s')") SIR_EOL, __func__);
     return true;
 }
 
 PLUGIN_EXPORT bool sir_plugin_write(sir_level level, const char* message) {
-    (void)printf("\t" DGRAY("plugin_sample (%s): level: %04"PRIx16", message: %s"),
```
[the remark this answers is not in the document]
johnsonjh: There was never a newline there before, so my script didn't add one.  I just added one.

bindings/ch/chsir.h
```diff
@@ -41,7 +41,7 @@ extern void* _Chsir_handle = dlopen("libsir.dl", RTLD_LAZY);
 int
 _chsir_early_abort(void) {
     (void)fprintf(_stderr, "Error: dlopen(): %s"
-                  "\n       Cannot get _Chsir_handle in %s:%d\n",
+                  "\x0d\x0a       Cannot get _Chsir_handle in %s:%d\x0d\x0a",
```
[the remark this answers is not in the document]
johnsonjh: This is in the C side of the binding, it doesn't include libsir until after this function, which is a Ch exec pragma abort handler, so you can't use SIR_EOL or any sir macros.  It also has to display correctly no matter the terminal mode, since the user might have set the mode to raw, so it should have the newline.

include/sir.hh
```diff
@@ -694,7 +694,7 @@ namespace sir
 
             bool write_out() {
                 for (auto it = _arr->rbegin(); it != _arr->rend(); it++) {
-                    if ((*it != '\0') && (*it == '\n')) {
+                    if ((*it != '\0') && (*it == '\x0a')) {
```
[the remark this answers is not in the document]
johnsonjh: Automated

include/sir/config.h
```diff
@@ -69,10 +69,33 @@
 # endif
 
 /**
- * Define to use <CR><LF> line endings, otherwise use <LF>.
+ * Define the system carriage return (CR) character.
  */
-# if !defined(SIR_EOL_CRLF)
-#  undef SIR_EOL_CRLF
+# if !defined(SIR_EOL_CR)
+#  define SIR_EOL_CR "\x0d"
```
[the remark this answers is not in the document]
johnsonjh: Automated.

tests/tests++.cc
johnsonjh: Maybe, but I'm not concerned enough to change the test harness here just because that code isn't getting tested. (And we know it works, and we've seen it "work" when we had a failing test.)

tests/tests.c
johnsonjh: It should be upper.  I'll fix.

tests/tests.c
johnsonjh: Automated

src/sirinternal.c
johnsonjh: Done.

src/sirinternal.c
johnsonjh: I'd have to check, but POSIX says we should do it.

"PTHREAD_MUTEX_INITIALIZER can be used to initialize any mutexes that are statically allocated. The effect shall be equivalent to dynamic initialization by a call to thread_mutex_init() with parameter attr specified as NULL, except that no error check is performed."

src/sirplugins.c
johnsonjh: That's not what I did though.  There are over a hundred\*\*\* places that gcovr doesn't hit, but I didn't put exclusions them, only on the specific places where it's apparent that it is not possible to create a failure in any simple way at runtime.  

\*\*\* 

```
src/sirerrors.c                              105     104    99%   237
src/sirmutex.c                                31      30    96%   50
example/example.c                             49      46    93%   66,92,176
src/sirqueue.c                                65      62    95%   39,65,93
src/sirtextstyle.c                            94      91    96%   73,119,173
include/sir.hh                               257     254    98%   712,728,826
include/sir/helpers.h                         82      78    95%   123-124,382,418
tests/tests++.cc                             211     204    96%   116-122
tests/tests_shared.c                         183     175    95%   165,183,211-212,246,258,278,312
src/sirhelpers.c                             191     180    94%   48-49,51,59,88,92,294,310,316,354-355
src/sirthreadpool.c                          103      92    89%   47,55-57,64-66,73-75,138
src/sirfilesystem.c                          127     115    90%   46,136,152,180,247,264,267-269,430,436,625
src/sirfilecache.c                           329     315    95%   42,83,146-147,187,199,242,260,307,329,339-340,631,644
src/sirplugins.c                             198     178    89%   80,209,216,245,283,300,309,318,342,349,371,381,389,401,408,431,438,451,460,500
tests/tests.c                               1214    1190    98%   1452,1580,1589-1590,1708-1709,1718,1781,1786-1787,1801-1802,2219,2250-2252,2270,2357,2461,2467,2513-2514,2526-2527
src/sirinternal.c                            593     554    93%   171-172,176-177,199-200,236-237,245-246,286,434,440,615,687,864,870,912,919-920,924-925,956,963-964,968-969,1057,1064,1107,1111,1118,1152,1236,1241,1326,1368,1475,1481
------------------------------------------------------------------------------
```

src/sirtextstyle.c
johnsonjh: Will remove for now

src/sirplugins.c
johnsonjh: Will remove for now then.

StoneCypher:
why would you want dos line endings though

johnsonjh:
> why would you want dos line endings though

DPS8M operates the terminal in raw mode, so newline alone only advances the line but will not reset the position to the initial column.

johnsonjh:
DMD still causing hassles, updated the issue, will have to keep the workaround for now.

johnsonjh:
@aremmell OK gtg

StoneCypher:
the ... the 36-bit honeywell system?

tests/tests_shared.c
```diff
@@ -115,7 +115,7 @@ void print_usage_info(const sir_cl_arg* args, size_t num_args) {
             longest = len;
     }
 
-    (void)fprintf(stderr, "\n" WHITE("Usage:") "\n\n");
+    (void)fprintf(stderr, SIR_WHITE("%s  Usage:%s%s"), SIR_EOL, SIR_EOL, SIR_EOL);
```
aremmell, line 118: Don't pass `SIR_EOL` as a arg with %s specifier

tests/tests_shared.c
```diff
@@ -101,10 +101,10 @@ void print_test_list(const sir_test* tests, size_t num_tests) {
         }
 
         if ((i % 2) != 0 || i == num_tests - 1)
-            (void)printf("\n");
+            (void)printf("%s", SIR_EOL);
     }
 
-    (void)printf("\n");
+    (void)printf("%s", SIR_EOL);
```
aremmell, line 107: Don't pass `SIR_EOL` as a arg with %s specifier

tests/tests_shared.c
```diff
@@ -101,10 +101,10 @@ void print_test_list(const sir_test* tests, size_t num_tests) {
         }
 
         if ((i % 2) != 0 || i == num_tests - 1)
-            (void)printf("\n");
+            (void)printf("%s", SIR_EOL);
```
aremmell, line 104: Don't pass `SIR_EOL` as a arg with %s specifier

tests/tests_shared.c
```diff
@@ -125,27 +125,27 @@ void print_usage_info(const sir_cl_arg* args, size_t num_args) {
             for (size_t n = len; n < longest; n++)
                 (void)fprintf(stderr, " ");
 
-        (void)fprintf(stderr, "%s%s%s\n", args[i].usage,
+        (void)fprintf(stderr, "%s%s%s" SIR_EOL, args[i].usage,
             strnlen(args[i].usage, SIR_CL_MAXUSAGE) > 0 ? " " : "", args[i].desc);
     }
 
-    (void)fprintf(stderr, "\n");
+    (void)fprintf(stderr, "%s", SIR_EOL);
```
aremmell, line 132: Don't pass `SIR_EOL` as a arg with %s specifier

tests/tests.c
```diff
@@ -148,7 +148,7 @@ int main(int argc, char** argv) {
         for (size_t t = 0; t < _sir_countof(sir_tests); t++)
             if (!sir_tests[t].pass)
                 print_failed_test(sir_tests[t].name);
-        (void)printf("\n");
+        (void)printf("%s", SIR_EOL);
```
aremmell, line 151: Don't pass `SIR_EOL` as a arg with %s specifier

aremmell (changes requested):

include/sir/config.h
```diff
@@ -69,10 +69,33 @@
 # endif
 
 /**
- * Define to use <CR><LF> line endings, otherwise use <LF>.
+ * Define the system carriage return (CR) character.
  */
-# if !defined(SIR_EOL_CRLF)
-#  undef SIR_EOL_CRLF
+# if !defined(SIR_EOL_CR)
```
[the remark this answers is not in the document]
johnsonjh: It using `SIR_USE_EOL_CRLF` that you can set as a make argument already.  I did it this way so you can override actual CR and LF characters used if you need to.

Like on some non-UNIX mainframes, they use `0x15` and `0x85` for these....

include/sir/config.h
```diff
@@ -373,8 +396,22 @@
  * data in any color mode (the largest possible sequence, which is:
  * `\x1b[a;fb;m;rrr;ggg;bbb;fb;m;rrr;ggg;bbbm`) plus a null terminator.
  */
-# if !defined(SIR_MAXSTYLE)
-#  define SIR_MAXSTYLE 43
+# if !defined(SIR_NO_TEXT_STYLING)
+#  if !defined(SIR_MAXSTYLE)
+#   if !defined(SIR_EOL_CRLF)
```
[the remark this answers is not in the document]
johnsonjh: Will fix.

include/sir/config.h
```diff
@@ -373,8 +396,22 @@
...
+#    define SIR_MAXSTYLE 43
+#   else
+#    define SIR_MAXSTYLE 44
+#   endif
+#  endif
+# else
+#  if !defined(SIR_MAXSTYLE)
+#   if !defined(SIR_EOL_CRLF)
```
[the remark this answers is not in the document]
johnsonjh: Fixed.

include/sir/platform.h
```diff
@@ -82,6 +89,10 @@
 #  define SANITIZE_SUPPRESS(str)
 # endif
 
+# if HAS_FEATURE(safe_stack) && !defined(SIR_NO_PLUGINS)
+#  error "linking DSO with SafeStack is unsupported; disable SafeStack or enable SIR_NO_PLUGINS"
```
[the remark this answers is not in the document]
johnsonjh: Fixed.

plugins/sample/plugin_sample.c
```diff
@@ -63,22 +63,22 @@ PLUGIN_EXPORT bool sir_plugin_query(sir_plugininfo* info) {
...
+    (void)printf("\t" SIR_DGRAY("plugin_sample (%s): level: %04"PRIx16", message: %s") SIR_EOL,
                  __func__, level, message);
     return true;
 }
 
 PLUGIN_EXPORT bool sir_plugin_cleanup(void) { //-V524
```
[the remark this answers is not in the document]
johnsonjh: There's no PVS suppression here - the only change was adding a newline that you requested.

tests/tests_shared.c
johnsonjh: Done.

tests/tests_shared.c
johnsonjh: Done.

tests/tests_shared.c
johnsonjh: Done.

tests/tests_shared.c
johnsonjh: Done.

tests/tests.c
johnsonjh: Done.

include/sir/config.h
```diff
@@ -69,10 +69,33 @@
 # endif
 
 /**
- * Define to use <CR><LF> line endings, otherwise use <LF>.
+ * Define the system carriage return (CR) character.
```
aremmell, line 72: `The carriage return (CR) character to use in the end of line sequence when SIR_USE_EOL_CRLF is defined.`

include/sir/config.h
```diff
@@ -69,10 +69,33 @@
...
  */
-# if !defined(SIR_EOL_CRLF)
-#  undef SIR_EOL_CRLF
+# if !defined(SIR_EOL_CR)
+#  define SIR_EOL_CR "\r"
+# endif
+
+/**
+ * Define the system line feed (LF) character.
```
aremmell, line 79: `The line feed (LF) character to use in the end of line sequence.`

include/sir/config.h
```diff
@@ -69,10 +69,33 @@
...
+ */
+# if !defined(SIR_EOL_LF)
+#  define SIR_EOL_LF "\n"
+# endif
+
+/**
+ * Define to have libsir use CRLF line endings, otherwise use LF.
+ */
+# if !defined(SIR_USE_EOL_CRLF)
```
aremmell, line 88: This is a NOOP; should just be documented in `docs/sources/config.md`

include/sir/config.h
```diff
@@ -69,10 +69,33 @@
...
+#  undef SIR_USE_EOL_CRLF
+# endif
+
+/**
+ * Define the end of line sequence.
```
aremmell, line 93: `The end of line sequence. If SIR_USE_EOL_CRLF is defined, the sequence will be SIR_EOL_CR + SIR_EOL_LF; otherwise just SIR_EOL_LF.`

include/sir/platform.h
```diff
@@ -82,6 +89,10 @@
 #  define SANITIZE_SUPPRESS(str)
 # endif
 
+# if HAS_FEATURE(safe_stack) && !defined(SIR_NO_PLUGINS)
+#  error "linking dynamically shared objects with SafeStack is unsupported; disable SafeStack or enable SIR_NO_PLUGINS"
```
aremmell, line 93: Now that I think of it, "linking dynamically shared objects" is redudnant (dynamic and shared imply each other), and if you dynamically load them it's technically not "linking." I would put "loading libraries dynamically while also using SafeStack is..."

tests/tests_shared.c
```diff
@@ -33,39 +33,39 @@
 #include "tests_shared.h"
 
 void print_intro(size_t tgt_tests) {
-    (void)printf(WHITEB("\n" ULINE("libsir") " %s (%s) running %zu %s...") "\n",
+    (void)printf(SIR_WHITEB(SIR_EOL SIR_ULINE("libsir") " %s (%s) running %zu %s...") SIR_EOL,
```
aremmell, line 36: SIR_EOL before SIR_WHITEB

aremmell (changes requested):
Sorry, couple last things

include/sir/platform.h
johnsonjh: I just used the terminology Clang uses in the documentation:

> SafeStack supports linking statically modules that are compiled with and without SafeStack. An executable compiled with SafeStack can load dynamic libraries that are not compiled with SafeStack. At the moment, compiling dynamic libraries with SafeStack is not supported.

> Linking a DSO with SafeStack is not currently supported.

johnsonjh: I can still change it if you want.

include/sir/config.h
johnsonjh: Done.

include/sir/config.h
johnsonjh: Done.

include/sir/config.h
johnsonjh: Removed from here and added to docs.

tests/tests_shared.c
johnsonjh: Good catch.

johnsonjh: Done.

aremmell (dismissed):

aremmell (approved):

PR #413, opened 2024-04-09
Silence a warning from the REUSE linter; update tested compilers

johnsonjh:
* Silence a warning from the REUSE linter
* While here, update tested compilers


johnsonjh (commented):
👍🏻

aremmell (approved):

PR #419, opened 2024-04-20
Add a suppression for semgrep

johnsonjh:
* Rule: cpp.lang.security.memory.unvalidated-array-index.unvalidated-array-index
* Finding Message: Externally controlled data is used to index a fixed-size array, `tmp`. This could lead to an out-of-bounds read or write, triggering memory corruption.
* https://gitlab.com/libsir/libsir/-/jobs/6733408769#L76


johnsonjh:
* https://gitlab.com/libsir/libsir/-/jobs/6733408769#L76

PR #438, opened 2024-08-20
Improve malloc debugging on OpenBSD, NetBSD, and FreeBSD

johnsonjh:


tests/tests.c
```diff
@@ -87,6 +87,8 @@ static const sir_cl_arg cl_args[] = {
 
 static sir_cl_config cl_cfg = {0};
 
+#include "tests_malloc_bsd.h"
```
aremmell, line 90: Plz move above to where other include statement is if possible

tests/tests_malloc_bsd.h
```diff
@@ -0,0 +1,55 @@
...
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * -----------------------------------------------------------------------------
+ */
+
+#ifndef _SIR_TESTS_MALLOC_BSD_H_INCLUDED
+# define _SIR_TESTS_MALLOC_BSD_H_INCLUDED
+
+# if defined(__FreeBSD__) && defined(DEBUG)
```
aremmell, line 35: I trust you know what you're doing here, cuz I sure don't

aremmell (changes requested):

tests/tests.c
johnsonjh: These need to be above the main statement, because malloc config is already initialized once main is called.

tests/tests_malloc_bsd.h
johnsonjh: I tested it and made sure it works 👍🏻

tests/tests.c
aremmell: Right... I meant move it up to 36 w the other include

aremmell (commented):
See reply re: include

tests/tests.c
johnsonjh: Oh alright

aremmell (approved):

PR #441, opened 2024-11-07
Release libsir 2.2.5

johnsonjh:
* Release libsir 2.2.5


include/sir.h
```diff
@@ -797,7 +797,7 @@ bool sir_syslogcat(const char* category);
  * **Example:**
  *
  * ~~~
- * 2.2.5-dev
```
aremmell, line 800: I think this is a grep mistake, unless we have been changing this comment in past releases...

aremmell (approved):

include/sir.h
johnsonjh: We haven't.  I'll change it back :)

Fixed in ef8d4013999d4f5985fe1814964f1d8eed929bc6

aremmell (approved):

PR #491, opened 2026-05-09
Update all CI tooling and regenerate all container images, etc.

johnsonjh:
Updated base CI container image to Fedora 43, rebuilt all downstream containers and got CI green again.


.gitlab-ci/msvc/msvc-wine/patches/Common7/Tools/vsdevcmd/core/winsdk.bat.patch
```diff
@@ -0,0 +1,74 @@
+#
+# Copyright (c) 2024 Huang Qinjin
```
aremmell, line 2: Are we going to need REUSE tags for these files?

src/sirfilecache.c
aremmell, line 244: Remove this `else` (I have this staged for PR on another branch already); hopefully this fixes the V7576 and we can remove that comment.

It is correct to remove the `if (sequence > 0)` because if it doesn't get incremented, we break out of the loop.

src/sirhelpers.c
```diff
@@ -343,12 +343,13 @@ char* _sir_strremove(char *str, const char *sub) {
     if (!sub)
         return str;
 
-    char* r;
+    const char* r;
+    const char* p;
```
aremmell, line 347: This will cause a new warning: "`p` can be declared in a smaller scope" - should remain where it was before, I'm pretty sure. Why did you move it up?

.lint.sh
```diff
@@ -378,15 +377,13 @@ test_gccextra()
         env CC="${CCACHE:-env} gcc"
             CXX="${CCACHE:-env} g++"
             CFLAGS="-Werror
-                    -Wbad-function-cast
```
aremmell, line 381: Did we fix this in the CI pipeline as well?

aremmell (changes requested):
I think there are some changes to be made still

.lint.sh
aremmell: (or is this the only source of the warnings in the CI logs)?

.gitlab-ci/msvc/msvc-wine/patches/Common7/Tools/vsdevcmd/core/winsdk.bat.patch
johnsonjh: No, covered by our existing REUSE toml config

src/sirhelpers.c
johnsonjh: No, this is required to avoid a warning that was triggered by another compiler after the glibc header updates - it doesn't produce any additional warnings, I tested it!

.lint.sh
johnsonjh: It fixes both!

src/sirfilecache.c
johnsonjh: Lets do it in the other PR branch then - I'd like to keep this one the minimum needed to get the CI working and this does it.  Especially since we are going to squash down to a single commit - OK with you?

aremmell (approved):

